PR #7, opened 2024-11-10
Add SQLite Backend

v0lkan:
This is an initial attempt to create a pluggable backing store.

Our first backing store candidate is SQLite; we'll have a file backing store, a bolt db backing store, and an s3 backing store later down the line.


v0lkan:
Looking good for an initial iteration.

Landing.

Will amend some more changes in a separate PR.

PR #38, opened 2024-11-17
refactor(app): spike-18 Should return an error store.Get, store.Delete

sahinakyol:
The methods that are mentioned in title should return an error for better audit tracking.


app/nexus/internal/state/state.go
```diff
@@ -130,12 +130,16 @@ func UpsertSecret(path string, values map[string]string) {
...
-	kv.Delete(path, versions)
 	kvMu.Unlock()
 
-	persist.AsyncPersistSecret(kv, path)
+	err := kv.Delete(path, versions)
+	if err != nil {
+		return err
+	} // TODO ? THIS RETURN BREAKS THE CODE FLOW ASK IS IT CORRECT ?
+
+	return persist.AsyncPersistSecret(kv, path)
```
v0lkan: This is an async operation; it should not be part of the return.

`persist.AsyncPersistSecret()` should NOT have a return value as it is a background operation.

app/nexus/internal/state/state.go
```diff
@@ -130,12 +130,16 @@ func UpsertSecret(path string, values map[string]string) {
...
+func DeleteSecret(path string, versions []int) error {
 	kvMu.Lock()
-	kv.Delete(path, versions)
 	kvMu.Unlock()
 
-	persist.AsyncPersistSecret(kv, path)
+	err := kv.Delete(path, versions)
+	if err != nil {
+		return err
+	} // TODO ? THIS RETURN BREAKS THE CODE FLOW ASK IS IT CORRECT ?
```
v0lkan: returning an `error` here is okay; but we also need to handle the error in where it is consumed:

```
	state.DeleteSecret(path, versions)
	log.Log().Info("routeDeleteSecret", "msg", "Secret deleted")
```

^ this section will need to catch that error and log if there is an error.

app/nexus/internal/route/delete.go
```diff
@@ -73,8 +74,12 @@ func routeDeleteSecret(
...
 	}
 
-	state.DeleteSecret(path, versions)
-	log.Log().Info("routeDeleteSecret", "msg", "Secret deleted")
+	err := state.DeleteSecret(path, versions)
+	if errors.Is(err, store.ErrSecretNotFound) {
+		log.Log().Warn("routeDeleteSecret", "msg", err.Error())
+	} else {
+		log.Log().Info("routeDeleteSecret", "msg", "Secret deleted")
+	}
```
v0lkan: +1

app/nexus/internal/route/secret_get.go
```diff
@@ -85,9 +86,13 @@ func routeGetSecret(
...
-	secret, exists := state.GetSecret(path, version)
-	if !exists {
-		log.Log().Info("routeGetSecret", "msg", "Secret not found")
+	secret, err := state.GetSecret(path, version)
+	if err != nil {
+		if errors.Is(err, store.ErrSecretNotFound) {
+			log.Log().Warn("routeGetSecret", "msg", err.Error())
+		} else if errors.Is(err, store.ErrSecretSoftDeleted) {
+			log.Log().Warn("routeGetSecret", "msg", err.Error())
+		}
```
v0lkan: We are logging the same thing; this comparison don't seem to add any value.

Consider using `errors.Join()` to add additional context to the message.

---

Also `.Warn()` might be too harsh for a 404-like "secret not found" error; `log.Info()` might be better.

app/nexus/internal/route/secret_get.go
```diff
@@ -96,8 +101,10 @@ func routeGetSecret(
 		}
 
 		net.Respond(http.StatusNotFound, responseBody, w)
-		log.Log().Info("routeGetSecret", "msg", "not found")
+		log.Log().Warn("routeGetSecret", "msg", "not found")
```
v0lkan: same. let's leave it as `Info()`; -- might change my mind later :D.

app/nexus/internal/state/state.go
v0lkan: ^ I take it back.. there is a `getrawsecret()` in that call which may return an error.

v0lkan: ^ I take that back. that method returning error is smelly.

it's better to log inside the method than returning an error.
will comment there too.

app/nexus/internal/state/persist/persist.go
```diff
@@ -37,11 +37,11 @@ var (
 // The function does not return any errors since it handles them internally
 // through logging. Cache failures are non-fatal as the KV store is considered
 // the authoritative data source.
-func AsyncPersistSecret(kv *store.KV, path string) {
+func AsyncPersistSecret(kv *store.KV, path string) error {
```
v0lkan: Instead of returning an error, suggest not returning the error but logging it instead.

app/nexus/internal/state/persist/persist.go
```diff
@@ -37,11 +37,11 @@ var (
...
 	be := Backend()
 
 	// Get the full secret for caching
-	secret := kv.GetRawSecret(path)
+	secret, err := kv.GetRawSecret(path)
```
v0lkan, line 44: Better to log the error here.

It could be possible that before we get secret to persist, someone else might have already deleted it because the following mutex does not lock over the async operation (and that's intentional).

```
	kvMu.Lock()
	kv.Put(path, values)
	kvMu.Unlock()

	persist.AsyncPersistSecret(kv, path)
```

So, not finding a secret is not an error; but more like a "warning" to be logged.

app/nexus/internal/state/state_test.go
```diff
@@ -0,0 +1,119 @@
+package state
+
+import (
+	"github.com/spiffe/spike/app/nexus/internal/state/store"
+	"testing"
+)
+
```
v0lkan, line 7: :pray: for the unit tests. Thanks a lot.

docs/quickstart.md
```diff
@@ -40,6 +40,9 @@ not a production-ready setup. For production, you should follow the
 Let's first build SPIRE from the source:
 
 ```bash
+echo 'export WORKSPACE="$HOME/-CHANGE_TO_DEVELOPMENT_DIRECTION-"' >> ~/.profile
```
v0lkan: do you mean "directory" instead of "direction"?

docs/quickstart.md
```diff
@@ -50,13 +53,13 @@ Add the SPIRE binaries to your PATH:
 ```bash
 # ~/.profile
 export PATH=$PATH:$WORKSPACE/spire/bin
+echo 'PATH=$PATH:$WORKSPACE/spire/bin' >> ~/.profile
```
v0lkan, line 56: +1.

app/nexus/internal/state/state.go
sahinakyol: I removed return then logged error in here

app/nexus/internal/state/persist/persist.go
sahinakyol: I left it as an info.

PR #49, opened 2024-11-24
Feat/17-retry mechanism

### LINKED ISSUE #17: implement exponential retries
This is a great library that we can leverage:

https://pkg.go.dev/github.com/cenkalti/backoff/v4#Retry

gurkanguray:
# Implement Exponential Retries

This PR implements exponential backoff retry mechanisms throughout the codebase using the [cenkalti/backoff](https://pkg.go.dev/github.com/cenkalti/backoff/v4#Retry) library, as suggested in #17.

## Changes

- Added new `internal/retry` package with:
  - Generic retry functionality with exponential backoff
  - Type-safe retry operations via `TypedRetrier`
  - Configurable retry parameters

- Implemented retries in critical operations:
  - Database operations in `persist` package
  - Admin token operations
  - Secret operations
  - Initialization operations

## Implementation Details

- Created standardized retry configurations with sensible defaults
- Added proper context handling and cancellation support
- Implemented comprehensive logging of retry attempts
- Maintained type safety through generic implementations

## Testing

- Added unit tests for retry mechanisms
- Verified retry behavior in database operations
- Tested timeout and cancellation scenarios

## Related Issues

Closes #17


internal/retry/retry.go
```diff
@@ -0,0 +1,122 @@
...
+) error {
+	b := r.newBackOff()
+	totalDuration := time.Duration(0)
+	return backoff.RetryNotify(
+		operation,
+		backoff.WithContext(b, ctx),
+		func(err error, duration time.Duration) {
+			totalDuration += duration
+			// log the error, duration and total duration
+			log.Log().Debug("Retrying operation after error", "error", err.Error(), "duration", duration, "total duration", totalDuration.String())
```
v0lkan: Instead of an explicit log, can we have a callback function (passed as an argument)? As in:

```
notify(err, duration, totalDuration)
```

That will make this module more generic.

internal/retry/retry.go
```diff
@@ -0,0 +1,122 @@
+//    \\ SPIKE: Secure your secrets with SPIFFE.
+//  \\\\\ Copyright 2024-present SPIKE contributors.
+// \\\\\\\ SPDX-License-Identifier: Apache-2.0
+
+package retry
```
v0lkan: Can we move this functionality to `./pkg/retry` instead?

I think it's a feature that can be consumed externally if other people need a similar functionality too.

v0lkan (changes requested):
I have two change requests:

* instead of logging use a callback function
* move the module under `pkg`

Other than that, it looks amazing 🤤 .

Thanks for your contributions 🎊 .

internal/retry/retry.go
gurkanguray: I added a notify option 🙌🏻

PR #60, opened 2024-12-24
Add input parameter sanitization

v0lkan:
Added
* input sanitization
* and policy enforcement to all API methods.


app/nexus/internal/route/acl/policy/create.go
```diff
@@ -79,34 +78,16 @@ func RoutePutPolicy(
 		return errors.ErrReadFailure
 	}
 
-	// TODO: sanitize
+	err := guardPutPolicyRequest(*request, w, r)
```
v0lkan, line 81: Poor man's aspect-oriented programming.

input validation and policy enforcement is a cross-cutting concern, so we are abstracting them in functions.

PR #66, opened 2024-12-30
feat: add test for list, put and undelete under pkg/store

yasinterol:
I want to add unit test test for list put and undelete under pkg/store

### DIFF
--- a/pkg/store/list_test.go
+++ b/pkg/store/list_test.go
@@ -1,0 +1,61 @@
+package store
+
+import "testing"
+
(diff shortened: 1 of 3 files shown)


v0lkan:
Thanks for your contribution @yasinterol -- tests around the kv store were much needed :pray: .

yasinterol:
thanks @v0lkan. I want to code test around the kv store as well. Thanks for your the review 🙏  one is greater than zero 😄

PR #73, opened 2025-01-02
Add basic lint, build and unit-test CI

strideynet:
I've kept this PR pretty basic to create a foundation for set of PRs that will add more CI/CD to SPIKE.

This PR contains the simplest CI, that on each commit to main, or to a PR open against main:

- A basic build of the SPIKE binary, to ensure that the changes have not broken the build.
- The unit test suite will be run. With the race detector, shuffling and coverage. A future PR can ship this coverage report somewhere more useful (e.g a comment on PRs, or a service like codecov)
- The Go Vet linter will be ran to catch simple issues.

You can see that this CI has run successfully on my fork: https://github.com/strideynet/spike/pull/1

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,5 +41,23 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: 'go.mod'
-      - name: Run Tests
+      - name: Unit Tests
         run: go test -v -race -shuffle=on -coverprofile=coverage.txt ./...
+  go-vet:
+    name: Go - Vet
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+      checks: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,63 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - name: Unit Tests
+        run: go test -race -shuffle=on -coverprofile=coverage.txt ./...
+  go-vet:
```
strideynet, line 46: Usually, I prefer to use `golangci-lint`, since this includes a wider set of linters. It also has a fairly nice support for GitHub Actions annotation mechanism. I held off adding it in this PR since I think it would be good to discuss this with other maintainers before adding it.

v0lkan: Yep, makes sense.

I'm okay with `golangci-lint` too. We can discuss with other maintainers as well.

Let's merge this, run it for a few times; and then we can amend as neeeded.

v0lkan:
LGTM.

Thanks four your contribution. 

Crossing fingers and landing :rocket: .

PR #77, opened 2025-01-03
Add GolangCI Lint

strideynet:
Closes: https://github.com/spiffe/spike/issues/19

You can view the CI pass on my fork: https://github.com/strideynet/spike/pull/2

I've stuck with the default set of linters for now, we may wish to add additional linters in the future.

### DIFF
--- a/app/keeper/internal/net/contribute.go
+++ b/app/keeper/internal/net/contribute.go
@@ -61,6 +61,9 @@ func Contribute(source *workloadapi.X509Source) {
 				Shard:    base64.StdEncoding.EncodeToString(contribution),
 			},
 		)
+		if err != nil {
+			panic("marshalling shard contribution request: " + err.Error())
+		}
 
 		_, err = net.Post(client, contributeUrl, md)
 		for err != nil {
--- a/app/keeper/internal/net/contribute.go
+++ b/app/keeper/internal/net/contribute.go
@@ -62,7 +62,9 @@ func Contribute(source *workloadapi.X509Source) {
 			},
 		)
 		if err != nil {
-			panic("marshalling shard contribution request: " + err.Error())
+			log.FatalLn(
+				"Failed to marshal shard contribution request: " + err.Error(),
+			)
 		}
 
 		_, err = net.Post(client, contributeUrl, md)
--- a/app/nexus/internal/net/cache/net.go
+++ b/app/nexus/internal/net/cache/net.go
@@ -6,6 +6,7 @@ package cache
 
 import (
 	"errors"
+	"fmt"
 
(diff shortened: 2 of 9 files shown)


app/keeper/internal/net/contribute.go
```diff
@@ -61,6 +61,9 @@ func Contribute(source *workloadapi.X509Source) {
 				Shard:    base64.StdEncoding.EncodeToString(contribution),
 			},
 		)
+		if err != nil {
+			panic("marshalling shard contribution request: " + err.Error())
```
strideynet: Some errors in this function lead to a Fatal and some lead to a Panic. Let me know if this should be a Fatal instead. I have a feeling we ought to refactor this to bubble the error up at somepoint regardless.

app/nexus/internal/net/cache/net.go
```diff
@@ -116,6 +117,11 @@ func FetchFromCache(source *workloadapi.X509Source) (string, error) {
 	}
 
 	data, err := net.Post(client, api.UrlKeeperRead(), md)
+	if err != nil {
+		return "", fmt.Errorf(
+			"FetchFromCache: failed to post request: %w", err,
```
strideynet, line 122: I see that other errors in this function don't propagate using error wrapping - is this intentional? If so, I can remove the wrapping from this error here.

app/keeper/internal/net/contribute.go
v0lkan: the current pattern is use log.FatalF or FatalLn instead of panicking.

That might be a remnant of a former proof-of-concept.

I have an action item to go through the entire codebase (for consistency); but haven't gotten there yet :).

app/nexus/internal/net/cache/net.go
v0lkan: My preference would be to wrap the error if we are pushing the error to the upper stack.

At least, that's the convention in the majority of the codebase.

v0lkan (approved):
Overall looking good.

I'll wait for a few days in case @strideynet wants to amend/change things.

Thanks for the automation 🙏 .

app/keeper/internal/net/contribute.go
strideynet: I will fix this today so we can merge.

strideynet: [842ffd2](https://github.com/spiffe/spike/pull/77/commits/842ffd283cfbd5690c08c6b34b7d1ac8b7452b35)

strideynet:
@v0lkan ready for merge.

v0lkan:
> @v0lkan ready for merge.

@strideynet will check it out tonight pacific (hopefully) :) .

PR #82, opened 2025-01-16
Nexus Recovery

v0lkan:
This PR contains the following flows

* Nexus bootstrapping and sending shards to keepers [1]
* Nexus creating a tombstone to mark that it has done (1)
* Nexus periodically re-sending shards to keepers [2]
* Nexus fetching shards from keepers if it crashes 

These cover majority of the DR use cases with an exception of the doomsdsay scenario where everything crashes without Nexus being able to sync its root key shards to keepers. [3]

- For (2), along with periodic shard sync, an additional improvement would be a keeper to call nexus if it does not have shard and it just started.
- For (3), we'll have a "securely export a shamir shards" functionality.
(both of these will be separate issues, and not part of this issue)


v0lkan:
The code needs some clean-up. 

Will add a comment here when it's mostly clean.

app/nexus/internal/env/backend.go
```diff
@@ -17,18 +17,24 @@ const (
...
+// TODO: add to docs, backing stores are considered untrusted as per the
+// security model of SPIKE; so even if you store it on a public place, you
+// don't lose much; but still, it's important to limit access to them.
+
 // BackendStoreType determines which storage backend type to use based on the
 // SPIKE_NEXUS_BACKEND_STORE environment variable. The value is
 // case-insensitive.
 //
 // Valid values are:
-//   - "file": Uses file-based storage
```
v0lkan, line 31: Removed "file" backing store, as "sqlite" is a file backing story anyway.

app/nexus/internal/env/backend.go
```diff
@@ -45,6 +51,6 @@ func BackendStoreType() StoreType {
 	case string(Memory):
 		return Memory
 	default:
-		return Memory
+		return Sqlite
```
v0lkan, line 54: Defaulting to SQLite.

We'll have a `--dev` flag that will disable keeper interaction and use an in-memory backing store (later down the line).

app/nexus/internal/env/database.go
```diff
@@ -97,6 +97,9 @@ func DatabaseConnMaxLifetimeSec() time.Duration {
 	return time.Hour
 }
 
+// TODO: ensure that all os.Getenv'ed env vars are documented in the readme.
```
v0lkan, line 100: I know there are a lot of `TODO:`s in the code; I'll take care of them in a follow-up PR.

PR #90, opened 2025-02-04
SDK update to fix a regression

v0lkan:
policy creation was failing due to overly strict spiffe id verification.

updated the regex in SDK and bumped SDK version here.


v0lkan:
which also means, unit tests (or integration tests) are not testing policy creation; but that's the story of another day :)

PR #91, opened 2025-02-11
Doomsday Recovery

v0lkan:
This is the first phase of doomsday recovery feature.

With this PR, in an unlikely scenario when SPIKE Nexus and >threshold SPIKE Keeper instances crash without having time to restore the system to a good configuration, then and operator can manually recover the system by using the `spike operator restore` command.

The feature works but right now it's tied with cotton threads
For example if the operator enters the same shard twice, the system will accept it without any due diligence, making recovery impossible.

I'll create a follow up PR(s) to address missing issues, including a documentation overhaul.


app/nexus/internal/initialization/initialization.go
```diff
@@ -31,8 +30,10 @@ import (
 func Initialize(source *workloadapi.X509Source) {
 	requireBootstrapping := env.BackendStoreType() == env.Sqlite
 	if requireBootstrapping {
-		bootstrap(source)
+		// Try bootstrapping in a loop.
+		go bootstrap(source)
```
v0lkan, line 34: Need to run this in a goroutine, otherwise the main SPIKE Nexus REST server would never start and it would be impossible to `spike operator recover` because there won't be any endpoint to listen to it.

And alternative option could be to make the `recover` functionality as part of the Nexus binary (instead of the Pilot binary), but I didn't want to give shell access to Nexus. -- an mTLS RESTful API is (arguably) more secure as we can guard it with SVIDs.

app/nexus/internal/initialization/recovery/recovery.go
```diff
@@ -38,20 +39,44 @@ import (
...
+		log.Log().Warn(fName, "msg", "You may need to manually bootstrap.")
+
 		log.Log().Info(fName, "msg", "Waiting for keepers to respond")
+
 		time.Sleep(5 * time.Second)
 	}
 }
 
+func RestoreBackingStoreUsingPilotShards(shards []string) {
+	firstShard := shards[0]
```
v0lkan, line 68: most of these methods assume 2 out of 3 shards and those numbers are hard coded.

I'll make it configurable in follow-up commit.

v0lkan: I wanted to make it work first and then work on make it "work better" and "work more flexible" :)

app/nexus/internal/route/base/route.go
```diff
@@ -35,6 +37,15 @@ func Route(
...
 		r.Method,
 		func(a net.SpikeNexusApiAction, p net.ApiUrl) net.Handler {
+			rk := state.RootKey()
+
+			emptyRootKey := rk == nil || len(rk) == 0
+			emergencyAction := p == net.SpikeNexusUrlOperatorRecover ||
+				p == net.SpikeNexusUrlOperatorRestore
+			if emptyRootKey && !emergencyAction {
+				return net.NotReady
+			}
```
v0lkan: if Nexus is locked the only actions available are `recover` and `restore`.

app/spike/internal/cmd/operator/restore.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/spiffe/go-spiffe/v2/workloadapi"
+	spike "github.com/spiffe/spike-sdk-go/api"
+	"golang.org/x/term"
+
+	"github.com/spiffe/spike/app/spike/internal/trust"
+	"github.com/spiffe/spike/internal/auth"
+	"github.com/spiffe/spike/internal/log"
+)
+
+func newOperatorRestoreCommand(
```
v0lkan, line 22: This and its "restore" counterpart require some cleanup.

I'll do that immediately after this PR.

app/spike/internal/cmd/policy/get.go
```diff
@@ -64,16 +66,20 @@ import (
...
 		Short: "Get policy details",
 		Args:  cobra.ExactArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
+			trust.Authenticate(spiffeId)
+
 			api := spike.NewWithSource(source)
 
-			policyID := args[0]
-			policy, err := api.GetPolicy(policyID)
+			policyId := args[0]
```
v0lkan, line 81: I know golang convention for stuff id is `stuffID`; but most of the code uses `stuffId` so I'm using that to be consistent here too.

app/spike/internal/cmd/secret/list.go
```diff
@@ -35,11 +37,15 @@ import (
...
 // Note: Requires an initialized SPIKE system and valid authentication
-func newSecretListCommand(source *workloadapi.X509Source) *cobra.Command {
+func newSecretListCommand(
+	source *workloadapi.X509Source, spiffeId string,
+) *cobra.Command {
 	var listCmd = &cobra.Command{
 		Use:   "list",
 		Short: "List all secret paths",
 		Run: func(cmd *cobra.Command, args []string) {
+			trust.Authenticate(spiffeId)
```
v0lkan, line 47: This is repeated a lot; I'm thinking of passing a higher-order function instead; but that will have to wait.

Issue #69, opened 2024-12-31
Exponential backoff for SPIKE Keepers' contribution endpoints


v0lkan:
This is how SPIKE keepers send their contribution

```
      _, err = net.Post(client, contributeUrl, md)
      for err != nil {
      time.Sleep(5 * time.Second)
      _, err = net.Post(client, contributeUrl, md)
      if err != nil {
      log.Log().Info("contribute",
      "msg", "Error sending contribution. Will retry",
      "err", err,
      )
      time.Sleep(5 * time.Second)
      }
      }
```

while this retry logic works, it is better to exponentially back off so that during a network partition we can give time for the system to breathe and recover.

There are examples of exponential backoff in the source code.

The point of this issue is to align with those examples, rather than implementing a backoff algorithm from scratch.

dsykes16:
I'd be happy to take this on. I've managed to drop an `ExponentialRetrier` from `spike-sdk-go` into the Nexus initialization recovery that had a fixed 5 second day, but I'd like to get some tests on it given the existing lack of coverage and my lack of familiarity with the codebase and retry library used.

v0lkan:
Closing this one.

Thanks for the contribution 🙏 .

PR #94, opened 2025-02-14
use `ExponentialRetrier` for Nexus recovery

dsykes16:
Replaces the fixed five second retry delay for Nexus recovery with an `ExponentialRetrier` from `spike-sdk-go`. This is related to #69 .

### DIFF
--- a/app/nexus/internal/env/recovery.go
+++ b/app/nexus/internal/env/recovery.go
@@ -1,0 +1,25 @@
+//    \\ SPIKE: Secure your secrets with SPIFFE.
+//  \\\\\ Copyright 2024-present SPIKE contributors.
+// \\\\\\\ SPDX-License-Identifier: Apache-2.0
+
+package env
+
+import (
+	"os"
+	"time"
+)
+
+// RecoveryOperationTimeout returns the recovery timeout duration.
+// It reads from the SPIKE_NEXUS_RECOVERY_TIMEOUT environment variable
+//
+// If the environment variable is not set or is not a valid duration string
+// then it defaults to `0` (no timeout limit).
+func RecoveryOperationTimeout() time.Duration {
+	e := os.Getenv("SPIKE_NEXUS_RECOVERY_TIMEOUT")
+	if e != "" {
+		if d, err := time.ParseDuration(e); err == nil {
+			return d
+		}
+	}
+	return 0
+}
--- a/app/nexus/internal/initialization/recovery/recovery.go
+++ b/app/nexus/internal/initialization/recovery/recovery.go
@@ -1,9 +1,11 @@
 package recovery
 
 import (
+	"context"
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"net/url"
 	"time"
 
@@ -11,12 +13,17 @@ import (
 	"github.com/spiffe/spike-sdk-go/api/entity/v1/reqres"
 	"github.com/spiffe/spike-sdk-go/crypto"
 	network "github.com/spiffe/spike-sdk-go/net"
+	"github.com/spiffe/spike-sdk-go/retry"
 
 	"github.com/spiffe/spike/app/nexus/internal/env"
 	state "github.com/spiffe/spike/app/nexus/internal/state/base"
 	"github.com/spiffe/spike/internal/auth"
 	"github.com/spiffe/spike/internal/log"
 	"github.com/spiffe/spike/internal/net"
+)
+
+var (
+	ErrRecoveryRetry = errors.New("recovery failed; retrying")
 )
 
 // RecoverBackingStoreUsingKeeperShards iterates through keepers until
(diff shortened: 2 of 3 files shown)


v0lkan:
Good idea.

Thanks for your contribution :pray: .

app/nexus/internal/initialization/recovery/recovery.go
```diff
@@ -43,25 +52,35 @@ func RecoverBackingStoreUsingKeeperShards(source *workloadapi.X509Source) {
...
-	for {
+	ctx, cancel := context.WithCancel(
+		context.Background(),
+	)
+	defer cancel()
+
+	retrier := retry.NewExponentialRetrier(
+		retry.WithBackOffOptions(
+			retry.WithMaxInterval(5),
+			retry.WithMaxElapsedTime(env.RecoveryOperationTimeout()),
```
v0lkan, line 63: +1 for having it retry forever by default.

app/nexus/internal/env/recovery.go
```diff
@@ -0,0 +1,25 @@
...
+	"time"
+)
+
+// RecoveryOperationTimeout returns the recovery timeout duration.
+// It reads from the SPIKE_NEXUS_RECOVERY_TIMEOUT environment variable
+//
+// If the environment variable is not set or is not a valid duration string
+// then it defaults to `0` (no timeout limit).
+func RecoveryOperationTimeout() time.Duration {
+	e := os.Getenv("SPIKE_NEXUS_RECOVERY_TIMEOUT")
```
v0lkan, line 18: You'll need to add this to the "configuration" section of the documentation too.

https://spike.ist/getting-started/configuration/

Here: https://github.com/spiffe/spike/blob/main/docs-src/content/getting-started/configuration.md

app/nexus/internal/initialization/recovery/recovery.go
```diff
@@ -43,25 +52,35 @@ func RecoverBackingStoreUsingKeeperShards(source *workloadapi.X509Source) {
...
+		),
+	)
+
+	retrier.RetryWithBackoff(ctx, func() error {
```
v0lkan: We need to check the error return value as the linter recommends.

v0lkan (changes requested):
I've had a comments; other than that it looks good.

Once they are addressed, we can merge it.

Thanks for your contribution 🙏 .

dsykes16:
> Good idea.
> 
> Thanks for your contribution 🙏 .

Thanks, but I do believe it was your idea based on the hint to standardize the methods used for retries in SPIKE per #69, so I can't take credit for it. 

I'll get the docs updated, issues the linter identified fixed up, and make sure there are some tests on this so I can kick it out of draft status. Thank you for your time to review this and all the helpful feedback, especially the config docs I completely overlooked.

app/nexus/internal/initialization/recovery/recovery.go
dsykes16: I wanted to match the existing behavior as closely as possible aside from the initial retries being faster than 5 seconds. 

I also wasn't sure the max elapsed time should even be configurable, and that's the main reason I went ahead and opened this draft PR before I'd finished adding tests for everything. My initial thought was it may be desirable to have some limit after which the recovery times out and the application exits with an error code, but on Kubernetes that would probably be better achieved with startup and liveness probes rather than an application-specific configuration, and for anything running on less ephemeral infrastructure (e.g. long-lived VMs/bare-metal) I'm not sure an app crash/restart is beneficial in any scenario (maybe by forcing a config reload?).

@v0lkan, do you think this config option actually adds value, or would you prefer I drop it?

v0lkan: @dsykes16 I've had mixed feelings, but I still think it has its merits / use cases.

My reasoning is, although majority of users will likely use SPIKE in an cotainerized environment (hence liveness and readiness probes), SPIKE is "by design" SPIFFE-first; so it will run wherever SPIFFE can run. Which means, if a readiness or liveness probe is not available, the user might prefer to retry for a pre-defined amount of time to eventually give up.

I think, as long as that interval is infinite (i.e. it retries forever) by default, it has feature parity. And if someone overrides it, then we'd assume they do want that override.

Let's keep the configuration for now.

app/nexus/internal/initialization/recovery/recovery.go
dsykes16: I added an error check with a call to `log.Fatal` to exit w/ non-zero return code on a retrier error.

dsykes16:
I believe all the requested changes have been resolved. I really wanted to add unit and/or integration tests on this, but I honestly struggled to find a clean way to unit test it as-is and I'm not comfortable with any substantial refactoring on a project I've only barely contributed to.

v0lkan:
Thanks. 

I'll check this out at my earliest convenience (not likely today :) )

v0lkan:
For the record; 

```
retry.WithMaxInterval(5),
```

I realized this ought to have been `5 * time.Second` instead of just `5` 

The current code was trying to loop ever 5 nanoseconds, which was a tad faster then intended :) 

---

Fixed that with a follow-up commit.

Thanks again for your contributions @dsykes16 .

v0lkan:
That's also why we need unit tests… but we'll get there :). 

At least we publish coverage reports in the repo to see where we can improve wrt testing (i.e. literally everywhere :D ) .

PR #99, opened 2025-02-19
feat(docker): Add flexible multi-stage Dockerfile for application builds

### LINKED ISSUE #16: figure out containarization

csoylu:
**Dockerfile**
- Supports building different applications with configurable architecture
- Uses multi-stage builds for optimized image creation
- Includes distroless base and static image targets

Helm charts are to come in another PR

fixes #16

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,55 @@
 # One Dockerfile to rule them all
-# Example:
-# docker build --build-arg APP=nexus -t keeper:latest --target distroless-base .
-FROM golang:1.23.3 AS base-builder
+FROM --platform=$BUILDPLATFORM golang:1.23.3 AS builder
+
+ARG APP
+ARG CGOENABLED
+ARG BUILDPLATFORM
+ARG TARGETPLATFORM
+ARG TARGETOS
+ARG TARGETARCH
+
+ENV GOOS=$TARGETOS \
+    GOARCH=$TARGETARCH \
+    GOEXPERIMENT=boringcrypto \
+    CGO_ENABLED=$CGOENABLED
+
 WORKDIR /workspace
+
+# Install cross-compilation tools
+RUN apt-get update && apt-get install -y \
+    gcc-x86-64-linux-gnu \
+    g++-x86-64-linux-gnu \
+    gcc-aarch64-linux-gnu \
+    g++-aarch64-linux-gnu \
+    libc6-dev-arm64-cross \
+    libc6-dev-amd64-cross
+
+
+# Download dependencies first (better layer caching)
 COPY go.mod go.sum ./
 RUN go mod download
 
-FROM golang:1.23.3 AS builder
-WORKDIR /workspace
-COPY --from=base-builder /go/pkg/mod /go/pkg/mod
+# Copy the app source code
 COPY . .
 
-ARG APP
-ARG TARGETARCH
-ENV GOOS=linux \
-    GOARCH=$TARGETARCH
-RUN if [ "nexus" = "$APP" ]; then \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN echo "Building $APP for architecture: $TARGETARCH"
+RUN go build -o $APP /workspace/app/$APP/cmd/main.go
+
+FROM gcr.io/distroless/base AS distroless-base
+COPY --from=builder /workspace/$APP /binary
+ENTRYPOINT ["/binary"]
+
+FROM gcr.io/distroless/static AS distroless-static
+COPY --from=builder /workspace/$APP /binary
+ENTRYPOINT ["/binary"]
```
v0lkan, line 55: Since we have `CGO_ENABLED` (at least for Nexus), we will need a mac (virtual) machine to create all cross-platform builds... (otherwise, for example a debian builder will not find relevant tool-chain to link mac/darwin bindings to build the binaries for the target os)

which is fine with me, I already have one mac build vm.

it just means, we'll need 9 (3 mac arm, 3 linux arm, 3 linux x86) container registries, instead of 3 -- which is fine with me, since there'll be some "form" of release automation anyway.

v0lkan: Upon further discussion on other channels, there may be hope :)

I'll test things out and get back here :).

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+    elif [ "$TARGETARCH" = "arm64" ]; then \
+        CC=aarch64-linux-gnu-gcc go build -o $APP /workspace/app/$APP/cmd/main.go; \
+    fi
+
+
+# Target distroless base image for CGO_ENABLED apps
+# This image includes a basic runtime environment with libc and other minimal dependencies
+FROM gcr.io/distroless/base AS distroless-base
+COPY --from=builder /workspace/$APP /binary
+ENTRYPOINT ["/binary"]
```
v0lkan, line 49: better to name the binary as `nexus`, `spike`, or `keeper` instead of `binary`.

Because we'll likely shell into some of those, similar to this:

```bash
kubectl exec $podName -- spike secret list
```

There `binary secret list` would look a bit weird.

v0lkan:
Merging this in as

1. nobody is using it
2. we don't even have a registry
3. but it's still a good starting point
4. I may find some time to iterate on it
5. @csoylu will likely be busy for the next few weeks.

Dockerfile
kfox1111: At least for Docker, there isn't a container runtime for mac. containers on mac are linux based.

Dockerfile
csoylu: we cannot use variables in ENTRYPOINT so when named to "nexus" for example we cant set the entrypoint to ["/nexus"] sinc we cant get $APP arg to work in entrypoint

kfox1111: May make sense then to have 3 separate Dockerfiles.

v0lkan: I am on the "separatre docker files" train too. -- that's how we ended up in VSecM too. -- each binary having its odnw dockerfile.

csoylu: put them all in a folder called kubernetes? where we would also have helms?

v0lkan: Example: https://github.com/vmware-tanzu/secrets-manager/tree/main/dockerfiles/vsecm-ist-fips

It's a bit repetitive but also "[A little copying is better than a little dependency](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)".

csoylu: i created an issu

https://github.com/spiffe/spike/issues/103

PR #102, opened 2025-03-02
Rehydrate secrets list from backing store

v0lkan:
If SPIKE Nexus crashes, it tries to re-seed the memory from the backing store, provided it's able to recover the root key.


csoylu (approved):

app/nexus/internal/initialization/recovery/recovery.go
```diff
@@ -163,6 +164,37 @@ func RecoverBackingStoreUsingKeeperShards(source *workloadapi.X509Source) {
...
+
+	secrets := persist.ReadAllSecrets()
+	if len(secrets) == 0 {
+		return
+	}
+
+	state.ImportSecrets(secrets)
+
+	log.Log().Info(fName, "msg", "HydrateMemoryFromBackingStore: secrets loaded")
+}
```
v0lkan, line 196: This is the majority of the magic.

`ImportSecrets()` is a new method introduced to the Go SDK.

PR #112, opened 2025-03-06
fix: Improve domain check for cross-platform compatibility in hack/start.sh

marikann:
# Fixed Domain Resolution Issue in WSL Environment

When running `make start` in a WSL environment, the existing `check_domain` function was unable to detect the `spire.spike.ist` domain record in the hosts file. However, using the `getent` command successfully resolved the domain.

With this PR, I improved the domain resolution mechanism:

1. First attempts resolution using the `dig` command
2. If unsuccessful, tries resolution using `getent`
3. For systems without `getent` (e.g., macOS), directly checks the hosts file

This change ensures more reliable script execution across different platforms (Linux, WSL, macOS, etc.).

### DIFF
--- a/hack/start.sh
+++ b/hack/start.sh
@@ -24,16 +24,33 @@
 SPIRE_SERVER_DOMAIN="spire.spike.ist"
 
 check_domain() {
-    # Try to resolve the domain and ensure we get an answer
+    # First try DNS resolution with dig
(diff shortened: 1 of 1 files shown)


v0lkan:
Thanks 🙏 — Merging this, as I'm going to test it in tomorrow's onboarding session anyway.

If it fails, it would be a good opportunity to live debug :) .

PR #114, opened 2025-03-08
create one dockerfile per app

### LINKED ISSUE #103: CI&CD: create one dockerfile per app
create one dockerfile for each up under /kubernetes and then in a separate issue, create hardened helm charts.

discussed in: https://github.com/spiffe/spike/pull/99

csoylu:
closes #103


v0lkan:
@csoylu I messages you on SPIFFE Slack too; repeating here just in case:

Since you are working on GitHub actions, it would be amazing if you can fix the Go Lint action too. 

It's giving a version mismatch error, and seeing red crosses at every merge is rather annoying :) .

csoylu:
Updated the linter version and we no longer have an error, the change that solves the error seems to be this one: https://github.com/golangci/golangci-lint/pull/5224

v0lkan (approved):
🌮 🐱 .

Issue #123, opened 2025-03-11
Add cache invalidation headers to API response


v0lkan:
It would be better security if SPIK Nexus sends no-cache headers as part of its RESTful API responses.

Something along these lines for ALL API responses:

```
w.Header().Set("Cache-Control", "no-store, no-cache, must-revalidate")
w.Header().Set("Pragma", "no-cache")
w.Header().Set("Expires", "0")
```

v0lkan:
This is done. Closing.

PR #124, opened 2025-03-11
Add cache invalidation headers to API responses (#123)

marikann:
## Add cache invalidation headers to API responses

This PR implements issue [#123](https://github.com/spiffe/spike/issues/123) by adding cache invalidation headers to all API responses from SPIKE Nexus.

### Changes:
- Added the following cache control headers to the `Respond` function:
  - `Cache-Control: no-store, no-cache, must-revalidate, private` (with "private" directive to prevent caching in shared caches)
  - `Pragma: no-cache`
  - `Expires: 0`
- Updated the function documentation to reflect these changes

### Why:
These headers prevent browsers and proxies from caching API responses, which enhances security by ensuring clients always receive fresh data and don't use potentially outdated sensitive information from cache. The "private" directive specifically prevents shared caches (like CDNs or proxies) from storing the response, restricting caching to the end user's browser only if at all.

### Implementation details:
The changes were made in the central `Respond` function within `response.go`, which is used by all API endpoints to send responses, ensuring consistent application of these headers across the entire API.

### DIFF
--- a/internal/net/response.go
+++ b/internal/net/response.go
@@ -55,9 +55,10 @@ func MarshalBody(res any, w http.ResponseWriter) []byte {
 
 // Respond writes a JSON response with the specified status code and body.
 //
-// This function sets the Content-Type header to application/json, writes the
-// provided status code, and sends the response body. Any errors during writing
-// are logged but not returned to the caller.
+// This function sets the Content-Type header to application/json, adds cache
+// invalidation headers (Cache-Control, Pragma, Expires), writes the provided 
+// status code, and sends the response body. Any errors during writing are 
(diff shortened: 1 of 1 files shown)


v0lkan:
Thanks for your contributions @marikann — looks good.

I'll let this marinate for a few days, and then merge it in :).

v0lkan:
Looking good.

I crossed my fingers and landed this :)

I'll do a preliminary test this weekend; and a full-blown regression before the next release cut.

We have some work items to put a 🎀 on before we cut the release though:
https://github.com/spiffe/spike/blob/main/jira.xml#L61-L67

PR #143, opened 2025-05-04
Stricter pattern matching for Secret and Policy Operations

v0lkan:
This PR introduces two predefined namespaces for policy-mangement purposes:

* `spike/system/acl`
* `spike/system/secrets`

These don't have any effect if the user is an administrator with a SPIKE Pilot SVID (SPIKE Pilot can CRUD policies and secrets with no restriction)

However, if we want to allow an arbitrary workload to, say, for example, list policies; then the workload will need to have a policy that defines LIST access for `spike/system/acl`.

We need this special `spike/system/secrets` path because the meta information about all secrets do not reside in a path, it is more like a schema-level operation as in "list all the secret paths in the system" -- so we assume we maintain a list of paths (and maybe other meta information in the future, too) under the pseudo `spike/system/secrets` paths, for convenience.

For policies, we have a similar issue. Policies don't have paths, so we consider all of them stay under the pseudo `spike/system/acl` path instead.

 because, technically, secrets (and policies)
do not live under a path--they themselves have paths, but Nexus REST API does not allow things like "list all secrets that are under a given path"

### DIFF
--- a/app/nexus/internal/route/acl/policy/create_intercept.go
+++ b/app/nexus/internal/route/acl/policy/create_intercept.go
@@ -43,9 +43,10 @@ func guardPutPolicyRequest(
 		return apiErr.ErrUnauthorized
 	}
 
+	// Request "write" access to the ACL system for the SPIFFE ID.
 	allowed := state.CheckAccess(
-		spiffeid.String(), "*",
-		[]data.PolicyPermission{data.PermissionSuper},
+		spiffeid.String(), "spike/system/acl",
+		[]data.PolicyPermission{data.PermissionWrite},
 	)
 	if !allowed {
 		responseBody := net.MarshalBody(reqres.PolicyCreateResponse{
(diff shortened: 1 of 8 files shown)


app/nexus/internal/route/acl/policy/create_intercept.go
```diff
@@ -43,9 +43,10 @@ func guardPutPolicyRequest(
 		return apiErr.ErrUnauthorized
 	}
 
+	// Request "write" access to the ACL system for the SPIFFE ID.
 	allowed := state.CheckAccess(
-		spiffeid.String(), "*",
-		[]data.PolicyPermission{data.PermissionSuper},
+		spiffeid.String(), "spike/system/acl",
```
v0lkan, line 48: These changes do not have any effect for the default mode of operation because if the request comes from SPIKE Pilot (with SPIKE Pilot SVID) then we assume that it can manage policies and secrets without requiring any policy attached to it.

The checks will only be useful if we want an arbitrary workload to create/read/list secrets or policies.

PR #145, opened 2025-05-07
KEEPER_URL format change.

v0lkan:


v0lkan:
> How are ids assigned? just by their position?

Right now yes. So if url1 is dead and we want url4, it would be `url4,url2,url3`

That's because the shamir library we use creates an incrementing ID (starting from 1) whenever we create a new shard.

There (in theory) should be ways to create shards with arbitrary ids (since it's just a coordinate—algebraically-speaking) but I haven't had a chance to look into that part of the code yet.

But it's a good idea to create an investigation user story for this.

I mean, if it's relatively easy to give the user freedom to choose an arbitrary number as a server index, than it would be much less error-prone, and much more devops-friendly.

I'll file an issue about it.
Thanks @kfox1111 .

Issue #152, opened 2025-05-21
spike secret get format


kfox1111:
spike secret get outputs in a human friendly way, but not a machine friendly way.

There needs to be a way to format the output. Needs to be able to get just the value of a single key without anything else in stdout, along with maybe the option of getting all key/values in json/yaml format.

PR #157, opened 2025-05-23
spike secret get format

kfox1111:
* Enables the output format to be specified (yaml, json, plain)
* Only the data will be placed on stdout.
* On error, an error code will be returned
* A single value can be fetched

Fixes: https://github.com/spiffe/spike/issues/152


app/spike/internal/cmd/secret/get.go
```diff
@@ -45,46 +48,92 @@ func newSecretGetCommand(
...
+				return fmt.Errorf("failure reading secret: %v", err.Error())
 			}
 
 			if secret == nil {
-				fmt.Println("Secret not found.")
-				return
+				return fmt.Errorf("Secret not found")
 			}
 
 			d := secret.Data
```
v0lkan, line 89: This could panic if Data is nil.

Suggest adding:

```
if secret.Data == nil {
    return fmt.Errorf("secret has no data")
}
```

app/spike/internal/cmd/secret/get.go
```diff
@@ -45,46 +48,92 @@ func newSecretGetCommand(
...
+								b, err = yaml.Marshal(v)
+							} else {
+								b, err = json.Marshal(v)
+							}
+							found = true
+							break
+						}
+					}
+				}
+				if err != nil {
```
v0lkan, line 127: suggest adding context instead of sending plain error

```
if err != nil {
    return fmt.Errorf("failed to marshal data: %w", err)
}
```

v0lkan:
I have minor nits, but overall looks good. Thanks for the update Kevin :pray: 

@kfox1111 feel free to address the comments and merge it.

Cheers.

PR #165, opened 2025-06-01
Containerization

v0lkan:
Containerization


app/nexus/cmd/main.go
```diff
@@ -39,12 +39,20 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	fmt.Println("SPIFFE Endpoint: ", spiffe.EndpointSocket())
```
v0lkan: I probably need to remove these.

PR #176, opened 2025-06-15
fix orphaned spike processes by adding exec command to scripts

rohatcan:
When running the startup script (`hack/bare-metal/startup/start.sh`) and terminating with Ctrl+C, SPIKE processes (Keeper and Nexus) were not being properly cleaned up and continued running as orphaned processes. Because the PID trapped is the wrapper script's PID so actual processes are not killed.

I still don't know why it's not happening for SPIRE Server and SPIRE Agent binaries but adding `exec` to SPIKE start scripts fixed it for me.


kfox1111 (approved):
LGTM. Thanks! :)

v0lkan:
Landed. Thanks for your contribution.

Issue #140, opened 2025-05-02
`spike policy create` should optionally ingest a yaml file


v0lkan:
`spike policy create` uses command line arguments to create policies.

This can be uncomfortable for certain DevOps/automation workflows, especially when you want to store the policies in a Git(Ops) repo to be applied later.

Have an option to ingest and create policies from a yaml file.

Something like

```
spike policy create my-policy ./the-policy-file.yaml
```

PR #177, opened 2025-06-21
Using yaml file for policy creation

parlakisik:
Policy has been created by parameters in the cli command. Instead of parameters in the cli ,  yaml file can be used too

Signed-off-by: Murat Parlakisik parlakisik@gmail.com

Fixes: #140

### DIFF
--- a/app/spike/internal/cmd/policy/apply.go
+++ b/app/spike/internal/cmd/policy/apply.go
@@ -1,0 +1,195 @@
+//    \\ SPIKE: Secure your secrets with SPIFFE.
+//  \\\\\ Copyright 2024-present SPIKE contributors.
+// \\\\\\\ SPDX-License-Identifier: Apache-2.0
+
+package policy
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/spiffe/go-spiffe/v2/workloadapi"
+	spike "github.com/spiffe/spike-sdk-go/api"
+	"github.com/spiffe/spike/app/spike/internal/trust"
+)
+
+// normalizePath removes trailing slashes and ensures consistent path format
+func normalizePath(path string) string {
+	if path == "" {
+		return path
+	}
+
+	// Remove all trailing slashes except for root path
+	if path != "/" {
+		path = strings.TrimRight(path, "/")
+	}
+
+	return path
+}
+
+// newPolicyApplyCommand creates a new Cobra command for policy application.
+// It allows users to apply policies via the command line by specifying
+// the policy name, SPIFFE ID pattern, path pattern, and permissions either
+// through command line flags or by reading from a YAML file.
+//
+// The command uses upsert semantics - it will update an existing policy if one
+// exists with the same name, or create a new policy if it doesn't exist.
+//
+// The command requires an X509Source for SPIFFE authentication and validates
+// that the system is initialized before applying a policy.
+//
+// Parameters:
+//   - source: SPIFFE X.509 SVID source for authentication
+//   - spiffeId: The SPIFFE ID to authenticate with
+//
+// Returns:
+//   - *cobra.Command: Configured Cobra command for policy application
+//
+// Command flags:
+//   - --name: Name of the policy (required if not using --file)
+//   - --spiffeid: SPIFFE ID pattern for workload matching (required if not using --file)
+//   - --path: Path pattern for access control (required if not using --file)
+//   - --permissions: Comma-separated list of permissions (required if not using --file)
+//   - --file: Path to YAML file containing policy configuration
+//
+// Valid permissions:
+//   - read: Permission to read secrets
+//   - write: Permission to create, update, or delete secrets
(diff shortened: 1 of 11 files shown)


app/spike/internal/cmd/policy/create.go
```diff
@@ -28,39 +40,55 @@ import (
...
 //
 //	spike policy create \
 //	    --name "web-service-policy" \
 //	    --spiffeid "spiffe://example.org/web-service/*" \
 //	    --path "/api/v1/*" \
 //	    --permissions "read,write"
 //
+// Example usage with YAML file:
+//
+//	spike policy create --file policy.yaml
```
kfox1111: I personally am much more in favor of upserts then separate insert/update type apis...

What do you all think of a kubernetes style:
```
spike policy apply --file policy.yaml
```

or even...
```
spike apply --file policy.yaml
```

app/spike/internal/cmd/policy/create.go
```diff
@@ -28,39 +40,55 @@ import (
...
+//
+// Example YAML file structure:
+//
```
kfox1111: I've seen a few other projects adopt k8s style/compatable apis

```
apiVersion: spike.ist/v1
kind: Policy
metadata:
  name: web-service-policy
spec:
  spiffeid: spiffe://example.org/web-service/*
  path: /api/v1/*
  permissions:
    - read
    - write
```

Would allow for a `spike apply -f policy.yaml` like interface maybe allowing secret loading via apply as well at some point, along with managing the policies via k8s operator some day directly.

v0lkan:
Thanks for your contribution @parlakisik; and welcome to the project.

v0lkan:
@parlakisik 
thanks for your hard work; it's very valuable, and will make DevOps’ life tiny bit easier :) 

@kfox1111 @parlakisik 
This is fairly large change.
I will review it within the week and provide my feedback.

Thanks a lot.

app/spike/internal/cmd/policy/create.go
parlakisik: That make sense.
if we want to configure it in k8s , then we need to define crd definition .
My suggestion is to go that path in multiple PRs .
Otherwise I have to define crd and kubernetes test into same PR .

app/spike/internal/cmd/policy/create.go
parlakisik: Should i deprecate the `spike policy create` command ?
we shouldn't define two different cli command to manage same configuration ?

app/spike/internal/cmd/policy/create.go
kfox1111: Yeah. no reason to implement k8s support at this point. just thinking an api that could leave room to be used that way in the future would be nice

app/spike/internal/cmd/policy/create.go
kfox1111: Hmm... I guess it depends on if we're keeping the arg based api still. I think so?

Maybe @v0lkan can weigh in here

K8s does keep a `kubectl create configmap --from-literal=foo=bar` kind of api in addition to apply.

v0lkan:
Update:
@parlakisik @kfox1111 

I tested the PR locally and it works on my machine (TM) :).

If we…

* remove trailing slashes from paths
* make paths less API-like (instead of /api/v1/blah use secrets/db/password for example
* rename `spike policy create` to `spike policy apply`
* update public documentation as needed.
* make non-file-based policies use upsert semantics too 
which would mean, instead of checking for a name collision, we update if exists, create if not exists (i.e. change the logic around the following code)… 
```go
// Check if a policy with this name already exists
exists, err := checkPolicyNameExists(api, policy.Name)
```

then we'll be off to the races.

v0lkan:
@parlakisik I see two unused functions:

```
  Error: app/spike/internal/cmd/policy/create.go:93:6: func `newPolicyCreateCommand` is unused (unused)
  func newPolicyCreateCommand(
       ^
  Error: app/spike/internal/cmd/policy/validation.go:71:6: func `checkPolicyNameExists` is unused (unused)
```

I'll re-pull and test the rest.

parlakisik:
@v0lkan  dont look it to the code yet. I am re-writing now .

v0lkan:
@parlakisik 

apply does not honor upsert semantics for non-file policies:

```
➜  spike git:(pr-177) ✗ ./examples/consume-secrets/demo-create-policy.sh 
Policy 'workload-can-read-1' applied successfully
Policy 'workload-can-write' applied successfully
➜  spike git:(pr-177) ✗ ./examples/consume-secrets/demo-create-policy.sh
Error: post: Problem connecting to peer
Error: post: Problem connecting to peer
```

* first run policies applied
* second run got an error (well the error description is misleading, but that's a different issue)

---

Also, if it's not asking too much, can you replace all `spike policy create`s with `spike policy apply` across the project?

---

The file-based policy upserted without any issue.

v0lkan:
Landed!

PR #181, opened 2025-07-13
introduce SPIKE Lite.

v0lkan:
This PR introduces SPIKE Lite (a mode that delegates some of the business logic to an object storage layer (such as minio, or AWS S3).

There is also additional scripts for the federation demo.


v0lkan:
@kfox1111 you might find this merge specially interesting ;).

Issue #185, opened 2025-07-18
Implement `spike operator lock/unlock`


v0lkan:
Locking is an operation where

* The root key will be zeroed out from SPIKE Nexus’ memory.
* Any backing store operation will be rejected.
* The shard sync/retrieve from SPIKE Keeper instances will be blocked.
* The system will remain locked if SPIKE Nexus is killed and brought back.
* Only `spike operator unlock` will revert these operations.
* When unlocked, SPIKE Nexus will rehydrate the root key by talking to the SPIKE keeper instances.
* When locked, `spike operator recover` and `spike operator restore` operations will be disabled too.
* The `lock` and `unlock` commands will require operator privileges (hence a separate SPIFFE ID).

kfox1111:
For ha setups, it complicates things needing to use a shared filesystem for locks... (same issue with init).

Maybe locked should be the default state on init of a new cluster? maybe locked but init the keepers is a supported operation for initial bootstrapping? not sure... need to think about this

kfox1111:
or make the locked state be an independent plugin type. then you can have, for example a plugin that stores the locked state in a k8s configmap for synchronization.

v0lkan:
> https://github.com/spiffe/spike/issues/185#issuecomment-3104439948

I have bigger concerns; 
we should be locking Nexus; not Pilot :)

Locking Pilot is locking the window while keeping the door (nexus) wide open.

I will add more context tonight.

I agree about lockfiles too; it makes HA a nightmare: not an issue for pilot; but we are “not” locking pilot; and nexus will have a ha setup in some globulous future :)

v0lkan:
most of my comments here were supposed to be comments on this PR:

https://github.com/spiffe/spike/pull/195

I'll add my thoughts to the PR, hopefully tonight pacific.

Bu @huseyingulec, you can follow this chatter to gain some initial insights too.

Thanks a lot for your effort and contribution;
I'm sure we'll get to the bottom of this 🙏 .

PR #195, opened 2025-07-22
feat(lock): add lock/unlock commands 

huseyingulec:
This PR implements the lock/unlock functionality for SPIKE Nexus with the following changes:

- Added new lock and unlock CLI commands under the operator command, allowing authorized operators to lock and unlock SPIKE Nexus.

- Introduced a file-based locking mechanism (/var/lib/spike/locked) to persist lock state across restarts.

- Enforced lock condition globally in operator and secret commands, preventing execution while SPIKE Nexus is locked.

- Added AuthenticateLock function to validate SPIFFE IDs for lock/unlock commands, requiring the lock(watchdog) role.

- Created a shell script (spire-server-entry-lock-unlock-register.sh) to register the required SPIFFE ID with appropriate selectors.

- Updated SPIFFE ID validation helpers to support lock/unlock authorization checks.

This PR fixes issue #185.

Note: This is not the final version. Some build bugs were encountered, so this PR is opened as a draft for now.

### DIFF
--- a/app/spike/internal/cmd/secret/new.go
+++ b/app/spike/internal/cmd/secret/new.go
@@ -5,8 +5,10 @@
 package secret
 
 import (
+	"fmt"
 	"github.com/spf13/cobra"
 	"github.com/spiffe/go-spiffe/v2/workloadapi"
+	"github.com/spiffe/spike/internal/lock"
 )
 
 // NewSecretCommand creates a new Cobra command for managing secrets.
@@ -18,6 +20,12 @@ func NewSecretCommand(
 	cmd := &cobra.Command{
 		Use:   "secret",
 		Short: "Manage secrets",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+        if lock.IsLocked() { 
+            return fmt.Errorf("SPIKE is locked — please unlock before running this command")
+        }
+        return nil
+    },
 	}
 
 	// Add subcommands to the policy command
--- a/hack/bare-metal/entry/spire-server-entry-lock-unlock-register.sh
+++ b/hack/bare-metal/entry/spire-server-entry-lock-unlock-register.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+#    \\ SPIKE: Secure your secrets with SPIFFE. — https://spike.ist/
+#  \\\\\ Copyright 2024-present SPIKE contributors.
+# \\\\\\\ SPDX-License-Identifier: Apache-2.0
+
+# Set path to the SPIKE binary
+PILOT_PATH="$(pwd)/spike"
+PILOT_SHA=$(sha256sum "$PILOT_PATH" | cut -d' ' -f1)
+
+# Register a new SPIFFE ID for lock/unlock
+spire-server entry create \
+  -spiffeID "spiffe://spike.ist/spike/pilot/role/watchdog" \
+  -parentID "spiffe://spike.ist/spire-agent" \
+  -selector "unix:uid:$(id -u)" \
+  -selector "unix:path:$PILOT_PATH" \
+  -selector "unix:sha256:$PILOT_SHA" \
+  -ttl 3600
+
+# Wait for propagation
+echo "Waiting for SPIFFE entry to propagate..."
+sleep 5
+echo "Watchdog SPIFFE ID registered successfully!"
--- a/internal/lock/lock.go
+++ b/internal/lock/lock.go
@@ -1,0 +1,37 @@
+//    \\ SPIKE: Secure your secrets with SPIFFE. — https://spike.ist/
+//  \\\\\ Copyright 2024-present SPIKE contributors.
+// \\\\\\\ SPDX-License-Identifier: Apache-2.0
+package lock
+
+import (
+    "os"
+    "sync/atomic"
+)
+
+var locked atomic.Bool
+
+const lockFile = "/var/lib/spike/locked" 
+
+// LockFile is the file used to indicate that SPIKE is locked.
+func init() {
+    if _, err := os.Stat(lockFile); err == nil {
+        locked.Store(true)
+    }
+}
+
+// Lock sets the SPIKE lock by creating a lock file and marking it as locked.
+func Lock() error {
+    locked.Store(true)
+    return os.WriteFile(lockFile, []byte("locked"), 0600)
+}
+
+// Unlock clears the SPIKE lock by removing the lock file.
+func Unlock() error {
+    locked.Store(false)
+    return os.Remove(lockFile)
+}
+
+// IsLocked checks if SPIKE is currently locked.
+func IsLocked() bool {
+    return locked.Load()
+}
--- a/app/spike/internal/cmd/operator/lock.go
+++ b/app/spike/internal/cmd/operator/lock.go
@@ -1,0 +1,66 @@
+package operator
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/spiffe/spike/internal/lock"
+
+	"os"
+
+	"github.com/spiffe/spike/app/spike/internal/env"
+
+	"github.com/spiffe/spike-sdk-go/spiffeid"
+
+	"github.com/spiffe/spike/app/spike/internal/trust"
+)
+
+// newOperatorLockCommand creates a new cobra command for locking
+// SPIKE Nexus, disabling all operations until it is explicitly unlocked.
+//
+// This command can only be used by operators with the `watchdog` role.
+//
+// Parameters:
+//   - source *workloadapi.X509Source: The X.509 source for SPIFFE authentication.
+//   - spiffeId string: The SPIFFE ID of the caller for role-based access control.
+//
+// Returns:
+//   - *cobra.Command: A cobra command that implements the lock functionality.
+//
+// The command performs the following operations:
+//   - Verifies the caller has the 'lock' (watchdog) role, aborting otherwise.
+//   - Authenticates the lock request.
+//   - Sends the lock request to the SPIKE API.
+//   - Prints confirmation if successful.
+func newOperatorLockCommand(spiffeId string,
+) *cobra.Command {
+	return &cobra.Command{
+		Use:   "lock",
+		Short: "Lock SPIKE Nexus and disable all operations",
+		Run: func(cmd *cobra.Command, args []string) {
+			if !spiffeid.IsPilotLock(env.TrustRoot(), spiffeId) {
+				fmt.Println("")
+				fmt.Println("  You need to have a `lock` role to use this command.")
+				fmt.Println("  Please run ./hack/bare-metal/entry/spire-server-entry-lock-unlock-register.sh")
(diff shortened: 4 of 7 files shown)


app/spike/internal/cmd/secret/new.go
```diff
@@ -18,6 +20,12 @@ func NewSecretCommand(
 	cmd := &cobra.Command{
 		Use:   "secret",
 		Short: "Manage secrets",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+        if lock.IsLocked() { 
+            return fmt.Errorf("SPIKE is locked — please unlock before running this command")
+        }
+        return nil
```
v0lkan, line 27: Do not check for locks inside SPIKE Pilot.

Lock checks will be done within SPIKE Nexus, and we should not be using a lock file for this.

hack/bare-metal/entry/spire-server-entry-lock-unlock-register.sh
```diff
@@ -0,0 +1,23 @@
...
+PILOT_SHA=$(sha256sum "$PILOT_PATH" | cut -d' ' -f1)
+
+# Register a new SPIFFE ID for lock/unlock
+spire-server entry create \
+  -spiffeID "spiffe://spike.ist/spike/pilot/role/watchdog" \
+  -parentID "spiffe://spike.ist/spire-agent" \
+  -selector "unix:uid:$(id -u)" \
+  -selector "unix:path:$PILOT_PATH" \
+  -selector "unix:sha256:$PILOT_SHA" \
+  -ttl 3600
```
v0lkan, line 18: we will check for this SPIFFE ID at the relevant SPIKE Nexus API endpoints; we are not going to check this at SPIKE pilot.

internal/lock/lock.go
```diff
@@ -0,0 +1,37 @@
...
+package lock
+
+import (
+    "os"
+    "sync/atomic"
+)
+
+var locked atomic.Bool
+
+const lockFile = "/var/lib/spike/locked"
```
v0lkan, line 13: do not use a lock file.

The best place to keep the locked/unlocked state is at the SPIKE Nexus’ backing store.

This way if SPIKE Nexus restarts, it will still be aware that it is locked or unlocked.

The locking operation MUST ensure the following (copying from the issue description; ill expand on each item separately):

## The root key will be zeroed out from SPIKE Nexus’ memory.

When we call the API endpoint and execute a successful **lock** operation, the following function shall return `true`

```
// RootKeyZero checks if the root key contains only zero bytes.
//
// If the rot key is zero and SPIKE Nexus is not in "in memory" mode,
// then it means SPIKE Nexus has not been initialized yet, and any secret
// and policy management operation should be denied at the API level.
//
// Returns:
//   - bool: true if the root key contains only zeroes, false otherwise
func RootKeyZero() bool {
	rootKeyMu.RLock()
	defer rootKeyMu.RUnlock()

	for _, b := range rootKey[:] {
		if b != 0 {
			return false
		}
	}
	return true
}
```

## Any backing store operation will be rejected.

Once locked, the **only** valid API operation is **unlock** (and optionally **status** (that will show whether the system is locked or unlocked)

Locking is a **system-level** operation; it's not done at the SPIKE Pilot ""at all"".

Why? Because if you lock SPIKE Pilot, but all Nexus APIs are still functional then

* Any legitimiate workload can still create and retrieve secrets: When system is locked, this is not allowed: We do not want that.

A system administrator who cannot assign SVIDs, but can `curl` into SPIKE Pilot can cleverly overcome the pseudo-lock restrictions by:

* fetching SVID and certificates from SPIRE Server or SPIRE Agent
* shell into SPIKE Pilot
* directly execute `curl` commands with the fetched SVIDs/certificates.
* Hence bypassing any CLI guardrails.

## The shard sync/retrieve from SPIKE Keeper instances will be blocked.

SPIKE Nexus fetches and sends Shamir Sharts to SPIKE Nexus instances.

When locked, those operations shall be paused too.

## The system will remain locked if SPIKE Nexus is killed and brought back.

Deleting the SPIKE Nexus process, or in a Kubernetes deployment, killing the pod and bringing SPIKE Nexus back up will still keep SPIKE Nexus in a locked state.

A new table might be useful to keep this system state:

```sql
CREATE TABLE IF NOT EXISTS system_state (
    key TEXT PRIMARY KEY,
    value TEXT NOT NULL,
    updated_time DATETIME NOT NULL
);

-- Insert default unlocked state
INSERT OR IGNORE INTO system_state (key, value, updated_time) 
VALUES ('locked', 'false', CURRENT_TIMESTAMP);
```

## Only `spike operator unlock` will revert these operations.

We are ignoring the obvious directly updating the backing store action: That can remain as an undocumented workaround. The official way to unlock the **system** shall be through `spike operator unlock` which in turn calls a `/api/v1/operator/unlock` RESTful API endpoint on SPIKE Nexus.

The SPIKE Backing Store shall include the following methods (and their impelementations both for SQLLite and Lite backing store variants)

```
	// Lock locks the backend, preventing write operations
	Lock(ctx context.Context) error
	// Unlock unlocks the backend, allowing write operations
	Unlock(ctx context.Context) error
	// Locked checks if the backend is currently locked
	Locked(ctx context.Context) (bool, error)
```

For SPIKE Lite, we can keep these operations as "noop" provided we publish an ADR about the reasoning behind it. SPIKE Lite does not use backing store; it only uses the "encryt" and "decrypt" API endpoints, thus the "locking" responsibility falls onto the object-store policies and the cloud admin (minio, s3, and the like) — 

^ @kfox1111 any suggestion for that; I think, locking nexus for the Lite mode is counterproductive since when it's lite mode Nexus is simply an encryption as a service; there is not much danger in that.

## When unlocked, SPIKE Nexus will rehydrate the root key by talking to the SPIKE keeper instances.

That means we will restar the "paused" sync-with-the-SPIKE-keepers workflows.

## When locked, `spike operator recover` and `spike operator restore` operations will be disabled too.

Self-explanatory: When locked, the only thing you can do is to unlock, and check status.

## The lock and unlock commands will require operator privileges (hence a separate SPIFFE ID).

That we have created in the script: `"spiffe://spike.ist/spike/pilot/role/watchdog"`; we are just validating it at the wrong place (and time).

The validation SHALL happen at SPIKE Nexus, at the other end of the API.

I think that's more or less it.

----

@huseyingulec if you are joining this Saturday's onboarding session, we can discuss the details separately; otherwise, we can chat offline.

I do realize that this task is a tad heftier than a typical "good first issue"; however, I honestly believe it will teach you a lot about SPIFFE, SPIRE, SPIKE, Cryptography and zero-trust security.

So, may the odds "ever" be in your favor.

We will always be around to push this issue and eventually put a 🎀 on it.

kfox1111:
I'm good with not supporting locking on lite.

Locking may be hard though for a native s3 api or any non sql db really and maybe a performance hit on multinode sql... especially depending on how its implemented.

the behavior being proposed is something like:
* user issues lock
* all servers lock in some order
* user gets back ok, they are locked
* all calls are now guaranteed to be locked...

For this to function, there are at least a few ways of doing this....
1. The data has to be pushed out to some place shared, and every nexus instance has to check that shared state, with every api request.... this can be heavy weight to check on something that rarely is true. Scaling of nexus should work pretty well though. autoscaling works smoothly. the lock location may become a scaling issue....
2. the data is shared out, and the nexus instances check on it periodically. if it fails to read, it shuts down or goes unresponsive...  the locking client needs to set the flag and then wait for the max timeout in order to guarantee all nexus agents are locked. Locking is slower, and there is a chance a little network blips could cause issues. api overhead is nonexistent. Scaling of Nexus works pretty well. autoscaling works easily.
3. an inventory of nexus's must be kept. the locking client contacts each nexus and locks them individually. The client needs direct access though, which is kind of painful in a lb'ed setup unless has direct access... But there is a hard "that server is locked" feedback for each server to the client. once they are all flagged, a stateful flag also should be stored that the nexus's will read on startup to look at locked state for new instances. api overhead is nonexistent in this case. Scaling of nexus is harder. autoscaling is still possible, but also much harder
4. 3, but the lock call goes through nexus before it tracks what nexus instances exist and fan it out. client doesnt have to have direct access or knowlege of nexus instances.

A lot of complexity all the way around. One should ask, is the complexity of the feature worth it at all? If it is, then I'd think the rarety of the use of the feature should be strongly considered vs the impact it will have on HA. HA will be used a lot more frequently, so lockings affect on it should be carefully considered.

Without much thinking about it yet, I think I'm a bit partial to option 2.

v0lkan:
All valid concerns.

Well some can be mitigated by a "overlord-minion" configuration. One overlord nexus will connect to the db at all times, and other minions just do pass-through proxy.

But still, the "overlord" will need to check the db at every single operation.

Maybe that's not the right way.

What we "actually" want is to shutd down nexus entirely.

Those can be done in several other ways:

* For kubernetes: scale down the StatefulSet to zero replicas.
* For non-kuberneetes: just kill the process(es)
* Or if you don't want to go nuclear, you can always unregistered Nexus‘ registration entries (or clusterspiffeid, depending on whether bare-metal or k8s) from SPIRE Server. — that will automatically render all connections unauthenticated, and block/freeze the mTLS connection.

And even if we disregard the db-check hit for the overlord-minion setup: Distributed systems are hard; and to figure out "who is the overlord", we will need imnplement a consensus algoritm, or piggyback on something/some-tool that already uses a consensus algorithm.

And in no time we will reinvent Consul :D. [1]

Well, [1] is still something that needs to be sorted out for a HA setup, but there are easier (and more solid) ways to lock the system.

For example:
Assume only an uber-admin has access to the SPIRE Server (isolated in its own secure box). Then only the uber-admin can lock the system by removing the registration entries. — That's  secure.

Compared to our scenario; an attacker who cannot issue a `spike operator lock` (for whatever reason) can still find a way to access the backing store (which is considere unprotected as per SPIKE security model) and just flip the lock/unlock flag.

That's too much complexity. 

There is a saying in Turkish: The stone we throw is not worth the amount of frogs we frighten.

@huseyingulec @kfox1111 :

I think I'm going to close this issue and publish an ADR with the reasoning(s) discussed here.

A script like the following is waaay effective than any sophisticated rocket surgery that we'll do here:

```
SPIFFE_ID="spiffe://spike.ist/spike/nexus/etc/etc"

# Find the Entry ID for the given SPIFFE ID
ENTRY_ID=$(spire-server entry show --spiffeID "$SPIFFE_ID" | awk '/Entry ID/ {print $NF}')

# Check if an Entry ID was found
if [ -z "$ENTRY_ID" ]; then
    echo "No entry found for SPIFFE ID: $SPIFFE_ID"
    exit 1
fi

# Delete the entry using the Entry ID
if spire-server entry delete --entryID "$ENTRY_ID"; then
    echo "Successfully deleted entry with SPIFFE ID: $SPIFFE_ID"
else
    echo "Failed to delete entry with SPIFFE ID: $SPIFFE_ID"
    exit 1
fi
```

We are SPIFFFE-native; so we better embrace that I guess :).

v0lkan:
Closing.

Relevant ADR: https://spike.ist/architecture/adrs/adr-0023/

Thanks for your exploration @huseyingulec ❤️ .

PR #201, opened 2025-07-29
feat(policy): Add filtering to policy list command

marikann:
This change introduces the ability to filter policies in the
`spike policy list` command by either a path pattern or a
SPIFFE ID pattern.

- Added `--path` and `--spiffeid` flags to `spike policy list`.
- These two filtering options are mutually exclusive and cannot be used together.
- The backend (Nexus API) has been updated to support these filter parameters.
- The command documentation has been updated to include the new flags and usage examples.

### DIFF
--- a/app/nexus/internal/route/acl/policy/list.go
+++ b/app/nexus/internal/route/acl/policy/list.go
@@ -17,12 +17,13 @@ import (
 	"github.com/spiffe/spike/internal/net"
 )
 
-// RouteListPolicies handles HTTP requests to retrieve all existing policies.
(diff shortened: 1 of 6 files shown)


v0lkan:
There are some CI failures. — I'll check the branch locally when I find some time.

v0lkan:
looks like we need a SDK upgrade for this to pass CI. — will look at it and update here.

marikann:
yes I think so too

v0lkan:
There is already a newer SPIKE SDK version.

I'll merge this and fix anything that needs fixing locally.

`main` will be broken for a while.

PR #202, opened 2025-08-10
Update policy-related documentation.

v0lkan:
Fixed documentations issues related to how patterns are used in policies.


v0lkan:
Hey @marikann @parlakisik, I'm mentioning you both because you both touched similar parts of the code.

This PR fixes several misrepresentations around how policies are parsed in the system.

As a quick summary:

* There is no glob-matching for policies; older documentation had implications like "`secrets/db/*` matches all secrets that start with secrets/db"---that's incorrect because the expression means "match anything that start with secrets/db followed by zero-to-infinity forward slashes.
  * `secrets/db///////` `secrets/db/potato` `secrets/db` are all matches.
  * also `test/secrets/db` will match.

In short, spiffe id, and path patterns in policies are not globs, but regular expressions.

The only exception is a single asterisk `"*"`, which would mean "match everything regardless".

examples/policies/sample-policy.yaml
```diff
@@ -1,12 +1,17 @@
...
+# This demonstrates how to define a policy using YAML instead of command line
+# flags
 
 # Policy name - must be unique within the system
 name: "web-service-policy"
 
 # SPIFFE ID pattern for workload matching
 # This pattern determines which workloads this policy applies to
-spiffeid: "spiffe://example.org/web-service/*"
+spiffeid: "spiffe://example.org/web-service/"
```
v0lkan, line 14: Removed trailing asterisk from policy files and examples since it is misleading.

These are not path globs, and the meaning of the asterisk is "repeat the last / 0-infinity times".

What might be a better fit there could be something like:

`"^spiffe://example\.org/web-service/.*$`
( start with "spiffe://example.org/web-serice/" followed by one or many characters )

app/nexus/internal/route/acl/policy/create.go
```diff
@@ -43,7 +43,7 @@ import (
 //
 //	{
 //	    "name": "example-policy",
-//	    "spiffe_id_pattern": "spiffe://example.org/*/service",
+//	    "spiffe_id_pattern": "spiffe://example\.org/*/service",
```
v0lkan: pattern is a regular expression, so we must escape the dot.

app/nexus/internal/route/acl/policy/create.go
```diff
@@ -43,8 +43,8 @@ import (
 //
 //	{
 //	    "name": "example-policy",
-//	    "spiffe_id_pattern": "spiffe://example.org/*/service",
-//	    "path_pattern": "/api/*",
+//	    "spiffe_id_pattern": "spiffe://example\.org/.*/service",
+//	    "path_pattern": "^/api/.*",
```
v0lkan, line 47: escaping the . and updating matches to reflect its regex nature.

app/nexus/internal/route/acl/policy/read.go
```diff
@@ -50,8 +50,8 @@ import (
 //	    "policy": {
 //	        "id": "policy-123",
 //	        "name": "example-policy",
-//	        "spiffe_id_pattern": "spiffe://example.org/*/service",
-//	        "path_pattern": "/api/*",
+//	        "spiffe_id_pattern": "^spiffe://example\.org/.*/service",
+//	        "path_pattern": "^api/",
```
v0lkan, line 54: emphasis on regular expressions.

PR #203, opened 2025-08-10
Add stricter linting.

v0lkan:
Add stricter linting to align better with Go coding conventions.

### DIFF
--- a/app/keeper/internal/env/tls.go
+++ b/app/keeper/internal/env/tls.go
@@ -6,14 +6,14 @@ package env
 
 import "os"
 
-// TlsPort returns the TLS port for the Spike Keeper service.
+// TLSPort returns the TLS port for the Spike Keeper service.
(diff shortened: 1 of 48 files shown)


v0lkan:
Hey @kfox1111 ,

I added this section to test mk:

```text
audit: test
	go mod tidy -diff
	go mod verify
	test -z "$(shell gofmt -l .)"
	go vet ./...
	go run honnef.co/go/tools/cmd/staticcheck@latest -checks=all,-ST1000,-U1000 ./...
	go run golang.org/x/vuln/cmd/govulncheck@latest ./...
```

It arguably creates more "prettification" warnings rather than functional errors that may actually break the code.

So I'm not sure if it's worth adding it to the CI.

A part of me says, "consistency is good; let CI run the same set of checks," and another part of me says, "no need for additional noise, keep things simple".

What do you think?

PR #205, opened 2025-08-11
fixes spike pilot CLI to work in windows.

v0lkan:
fixes spike pilot CLI to work in Windows.

### DIFF
--- a/app/spike/internal/cmd/operator/restore.go
+++ b/app/spike/internal/cmd/operator/restore.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"strconv"
 	"strings"
-	"syscall"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


app/spike/internal/cmd/operator/restore.go
```diff
@@ -81,7 +80,7 @@ func newOperatorRestoreCommand(
 
 			fmt.Println("(your input will be hidden as you paste/type it)")
 			fmt.Print("Enter recovery shard: ")
-			shard, err := term.ReadPassword(syscall.Stdin)
+			shard, err := term.ReadPassword(int(os.Stdin.Fd()))
```
v0lkan, line 83: This change makes the file descriptor platform agnostic.

Otherwise, the binary was not compiling in Windows.

PR #206, opened 2025-08-13
Separate Bootstrap Logic into its own app

v0lkan:
This PR separates the bootstrapping flow into its own app, opening doors for a more predictable HA mode.


app/bootstrap/cmd/main.go
```diff
@@ -0,0 +1,41 @@
...
+	defer spiffe.CloseSource(src)
+
+	for keeperID, keeperAPIRoot := range env.Keepers() {
+		net.Post(
+			net.MTLSClient(src),
+			url.KeeperEndpoint(keeperAPIRoot),
+			net.Payload(state.KeeperShare(state.RootShares(), keeperID), keeperID),
+			keeperID,
+		)
+	}
```
v0lkan: This is the gist of the code:

1. Compute shards
2. for every keeper, send the keeper its own shard.

The next step would be to wait for Nexus to call keepers and initialize itself (and profit).

The change will simplify nexus code too, since it was relying on tombstone files and other heuristics to decide whether it needs to bootstrap or lock itself and wait for human intervention.

In this case
1. We run a bootstrap script (or k8s job) 
2. Ensure that the job succeeds
3. And we won't need to run the job ever again.

That will enable two SPIKE Keeper instances behind a load-balancer, using the same root keys
(ideally configured in an active/standby setup) without any zero-day state initialization/synchronization issues.

.github/workflows/build.yaml
```diff
@@ -13,29 +17,33 @@ jobs:
     runs-on: ${{ matrix.archmap[matrix.arch] }}
     strategy:
       matrix:
-        app: [pilot, keeper, nexus]
+        app: [pilot, keeper, nexus, demo, bootstrap]
```
v0lkan, line 20: Adding demo and bootstrap to the build matrix.

Demo can be used as an image to test SDK integration in e2e tests;
And Bootstrap is the new feature where we bootstrap SPIKE Nexus out-of-band to allow future HA capabilities.

.github/workflows/release.yaml
```diff
@@ -54,16 +61,9 @@ jobs:
...
-          ./hack/docker/build-push-sign.sh ${{ matrix.app }} ${{ matrix.arch }} ${VERSION} ${{ env.REGISTRY }} ${{ env.REPOSITORY }}
-
-#      - name: Sign the images with GitHub OIDC (Cosign)
-#        env:
-#          DIGEST: ${{ steps.build-and-load.outputs.digest }}
-#          TAGS: ${{ steps.meta.outputs.tags }}
-#        run: |
-#          echo "${TAGS}" | tr ',' '\n' | while read -r tag; do
-#            cosign sign --yes "${tag}@${DIGEST}"
-#          done
```
v0lkan, line 66: since we are not using this anymore, removed that.

Makefile
```diff
@@ -1,6 +1,6 @@
-#   \\
-#  \\\\ SPIKE: Secure your secrets with SPIFFE. — https://spike.ist/
-# \\\\\\
+#    \\ SPIKE: Secure your secrets with SPIFFE. — https://spike.ist/
+#  \\\\\ Copyright 2024-present SPIKE contributors.
+# \\\\\\\ SPDX-License-Identifier: Apache-2.0
```
v0lkan, line 3: some of these are global find/replaces to keep source code banners consistent.

app/bootstrap/cmd/main.go
```diff
@@ -0,0 +1,84 @@
...
+	flag.Parse()
+	if !*init {
+		fmt.Println("")
+		fmt.Println("Usage: bootstrap -init")
+		fmt.Println("")
+		os.Exit(1)
+		return
+	}
+
+	// Check if we should skip bootstrap (set by init container)
```
v0lkan, line 36: We may change this logic once we implement a `/status` endpoint for SPIKE Nexus.

app/bootstrap/cmd/main.go
```diff
@@ -0,0 +1,84 @@
...
+	if !svid.IsBootstrap(env.TrustRoot(), sv.ID.String()) {
+		log.Log().Error(
+			"Authenticate: You need a 'boostrap' SPIFFE ID to use this command.",
+		)
+		os.Exit(1)
+		return
+	}
+
+	log.Log().Info(
+		fName, "FIPS 140.3 enabled", fips140.Enabled(),
```
v0lkan, line 66: this returns true/false depending on how the binary is compiled. -- I modified the scripts to generate binaries where fips is enforced.

app/bootstrap/internal/state/state.go
```diff
@@ -0,0 +1,107 @@
...
+
+	log.Log().Info(fName, "t", t, "n", n)
+
+	// Create a secret from our 32-byte key:
+	rootSecret := g.NewScalar()
+
+	if err := rootSecret.UnmarshalBinary(rootKeySeed[:]); err != nil {
+		log.FatalLn(fName + ": Failed to unmarshal key: %v" + err.Error())
+		os.Exit(1)
+	}
```
v0lkan, line 51: There might be some code duplication since I copied these functions from Nexus into the `internal` modules of this app. -- Once we are sure things work, we can do a follow-up clean-up to consolidate common code and remove the code that is not used anymore.

app/nexus/internal/initialization/recovery/recovery.go
```diff
@@ -72,34 +71,28 @@ func RecoverBackingStoreUsingKeeperShards(source *workloadapi.X509Source) {
...
-			"message", "Waiting for keepers to respond")
+			"message", "Initialization unsuccessful. Will retry.",
+			"keepersSoFar", len(successfulKeeperShards),
+		)
 		return false, ErrRecoveryRetry
 	},
 		retry.WithBackOffOptions(
 			retry.WithMaxInterval(env.RecoveryOperationMaxInterval()),
-			retry.WithMaxElapsedTime(env.RecoveryOperationTimeout()),
+			retry.WithMaxElapsedTime(0), // Retry forever.
```
v0lkan, line 90: maybe we can add a configurable "bootsrap time out" as an env var later. For now, the bootsrap app will try to bootstrap the thing in an exponentially-backing-off loop until it succeeds.

app/nexus/internal/initialization/recovery/recovery.go
```diff
@@ -72,34 +71,28 @@ func RecoverBackingStoreUsingKeeperShards(source *workloadapi.X509Source) {
...
 		),
 	)
 
 	if err != nil {
-		log.Log().Warn("Recovery failed; timed out")
+		log.Log().Warn("Initialization failed; timed out")
```
v0lkan: This was part of a recovery flow; but now it is an initialization/bootsrapping logic; renamed the message to be more semantically correct.

app/nexus/internal/route/base/route.go
```diff
@@ -78,10 +78,9 @@ func Route(
 				return operator.RouteRestore
 			case a == url.ActionDefault && p == url.NexusOperatorRecover:
 				return operator.RouteRecover
-			// FIXME move constant
-			case a == url.ActionDefault && p == "/v1/cipher/encrypt" && !fullMode:
+			case a == url.ActionDefault && p == url.NexusCipherEncrypt && !fullMode:
 				return cipher.RouteEncrypt
-			case a == url.ActionDefault && p == "/v1/cipher/decrypt" && !fullMode:
+			case a == url.ActionDefault && p == url.NexusCipherDecrypt && !fullMode:
```
v0lkan, line 83: moved these to consts.

docs-src/content/development/local-deployment.md
```diff
@@ -121,14 +121,30 @@ registry:
...
+> 
+> One way to push container images to the Minikube cluster without adding them
+> to the local Minikube registry would be to run the following script:
+> 
+> ```bash
+> make k8s-load-images
+> ```
+> 
+> This will effectively have the same end-result as pushing the images to
+> the registry.
```
v0lkan, line 138: WSL is very good, and it has great Docker for Windows integration... until it starts giving weird network-related errors and freezing docker push'es for various reasons.

An alternate approach could be to install docker to the WSL linux locally and not use windows docker at all. 

But `make k8s-load-images` directly bypasses registry and saves the day anyway.

docs-src/content/getting-started/quickstart.md
```diff
@@ -109,24 +138,179 @@ helm upgrade --install spiffe spire \
...
-## Verifying SPIKE Deployment
+## Bootstrapping **SPIKE Nexus**
+
+To use **SPIKE Nexus**, we'll need to run a bootstrapper job that will
+seed it with a secure random root key.
+
+At the time of this writing, there is an ongoing work to automate this at
+**SPIFFE Helm Charts** upstream repo; however, until that work is merged and
+published, you'll need to create the following YAML file and apply it
+using `kubectl`.
```
v0lkan, line 149: We'll need to update this doc once Bootstrap lands to helm-charts.

docs-src/content/operations/release.md
```diff
@@ -40,13 +40,7 @@ Before every release:
...
 7. Run `./hack/cover.sh` to update and send the coverage report to the public 
    docs.
-8. Make sure version numbers in the code are updated to reflect the release:
-
-```go
-const SpikeNexusVersion = "0.4.2"
-const SpikePilotVersion = "0.4.2"
-const SpikeKeeperVersion = "0.4.2"
-````
+8. Make sure you update `./app/VERSION.txt` with the new version.
```
v0lkan, line 43: Along with many tidbidts introduced in this PR, I've also enabled Go Embed FS; so whenever we modify this VERSION.txt and trigger a build, all binaries will have a hard-coded version number in them. we won't have to update the source code. -- much easier to automate; much less error prone.

hack/k8s/spike-install.sh
```diff
@@ -53,24 +53,15 @@ echo "SPIKE namespaces:"
 kubectl get namespaces | grep spike || echo "No spike namespaces found"
 
 helm upgrade --install -n spire-mgmt spire-crds spire-crds \
-  --repo https://spiffe.github.io/helm-charts-hardened/ --create-namespace
+  --repo https://spiffe.github.io/helm-charts-hardened/ \
+  --version 0.26.1 --create-namespace
```
v0lkan, line 57: I pinned versions, just in case. but we should remember to update and test with newer versions too.

makefiles/Kubernetes.mk
```diff
@@ -34,10 +51,15 @@ docker-push:
 # For Multi-Cluster Federation Demo, DO NOT run `deploy-local`
 # Instead, see FederationDemo.mk for the remaining steps.
 
-# 6. (Single Cluster) Deploy SPIRE and SPIKE to the cluster.
+# DEPRECATED
```
v0lkan, line 54: Keeping this for a while, but will likely remove it.

PR #210, opened 2025-08-16
Sync CI pipeline with makefile targets and remove `golangci-lint` dependency 

mertssmnoglu:
This PR improves the development and CI setup of the spike by introducing consistent code quality checks.
And keeps the spike-sdk-go and spike code quality checks consistent. | https://github.com/spiffe/spike-sdk-go/pull/35

## Changes

- Removed `golangci-lint` from the dependencies, since it is only needed as a linting tool and not for build or test. We can separately run it.
- Extend audit to run golangci-lint too, (it's no longer calls `make test`).
- Created a new `ci` target, runs both `test` and `audit` targets.
- To sync make audit and the CI pipeline, I removed golangci-lint action and run checks with `make audit`.


mertssmnoglu:
I also have a couple of follow-up questions:

- Do you want me to align/sync the current `make test` target with the existing unit-test pipeline? If yes, which one?
- Since `make audit` is working and already covers code quality checks, do we still need the `hack/qa/lint-go.sh` script and the `make lint-go` target?

PR #212, opened 2025-08-17
Add spiffe cipher encrypt/decrypt cli commands

parlakisik:
First template to create cli commands

### DIFF
--- a/app/spike/internal/cmd/cipher/decrypt.go
+++ b/app/spike/internal/cmd/cipher/decrypt.go
@@ -1,0 +1,122 @@
+//    \\ SPIKE: Secure your secrets with SPIFFE. — https://spike.ist/
+//  \\\\ Copyright 2024-present SPIKE contributors.
+// \\\\\\ SPDX-License-Identifier: Apache-2.0
+
+package cipher
+
+import (
+	"encoding/base64"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+
+	"github.com/spf13/cobra"
+	"github.com/spiffe/go-spiffe/v2/workloadapi"
+	sdk "github.com/spiffe/spike-sdk-go/api"
+
+	"github.com/spiffe/spike/app/spike/internal/stdout"
+	"github.com/spiffe/spike/app/spike/internal/trust"
+)
+
+func newDecryptCommand(source *workloadapi.X509Source, SPIFFEID string) *cobra.Command {
+	var inFile string
+	var outFile string
+	var versionStr string
+	var nonceB64 string
+	var ciphertextB64 string
+	var algorithm string
+
+	cmd := &cobra.Command{
+		Use:   "decrypt",
+		Short: "Decrypt file or stdin via SPIKE Nexus",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			trust.AuthenticateForPilot(SPIFFEID)
+
+			jsonMode := versionStr != "" || nonceB64 != "" || ciphertextB64 != ""
+
+			var in io.ReadCloser
+			if !jsonMode {
+				if inFile != "" {
+					f, err := os.Open(inFile)
+					if err != nil {
+						return fmt.Errorf("failed to open input file: %w", err)
+					}
+					in = f
+				} else {
+					in = os.Stdin
+				}
+				defer func() {
+					if in != os.Stdin {
+						_ = in.Close()
+					}
+				}()
+			}
+
+			var out io.Writer
+			var outCloser io.Closer
+			if outFile != "" {
+				f, err := os.Create(outFile)
+				if err != nil {
+					return fmt.Errorf("failed to create output file: %w", err)
+				}
+				out = f
+				outCloser = f
+			} else {
+				out = os.Stdout
+			}
+			if outCloser != nil {
+				defer outCloser.Close()
+			}
+
+			api := sdk.NewWithSource(source)
+			if !jsonMode {
+				plaintext, err := api.CipherDecryptStream(in, "application/octet-stream")
+				if err != nil {
+					if err.Error() == "not ready" {
+						stdout.PrintNotReady()
+					}
+					return fmt.Errorf("failed to call decrypt endpoint: %w", err)
+				}
+				if _, err := out.Write(plaintext); err != nil {
+					return fmt.Errorf("failed to write output: %w", err)
+				}
+				return nil
+			}
+
+			// JSON mode
+			v, err := strconv.Atoi(versionStr)
+			if err != nil || v < 0 || v > 255 {
+				return fmt.Errorf("invalid --version, must be 0-255")
+			}
+			nonce, err := base64.StdEncoding.DecodeString(nonceB64)
+			if err != nil {
+				return fmt.Errorf("invalid --nonce base64: %w", err)
+			}
+			ciphertext, err := base64.StdEncoding.DecodeString(ciphertextB64)
+			if err != nil {
+				return fmt.Errorf("invalid --ciphertext base64: %w", err)
+			}
+			plaintext, err := api.CipherDecryptJSON(byte(v), nonce, ciphertext, algorithm)
+			if err != nil {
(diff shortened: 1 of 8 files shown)


app/spike/internal/cmd/cipher/client.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	transport := &http.Transport{
+		TLSClientConfig: &tls.Config{
+			Certificates: []tls.Certificate{tlsCert},
+			RootCAs:      roots,
+			MinVersion:   tls.VersionTLS12,
+		},
+	}
+	client := &http.Client{Transport: transport}
```
v0lkan: We should not need to go this low-level.

`go-spiffe` nicely abstracts SDVI/certificate retrieval for us.

Please consider leveraging `go-spiffe` 

An example can be found here: 
https://github.com/spiffe/spike-sdk-go/blob/main/api/internal/impl/api/secret/get.go#L51-L54

https://github.com/spiffe/spike-sdk-go/blob/main/api/internal/impl/api/secret/get.go#L11

The following (and maybe other) packages provide all the boilerplate we need, and have been battle-tested in SPIRE and SPIRE Controller Manager codebases:

```
	"github.com/spiffe/go-spiffe/v2/spiffeid"
	"github.com/spiffe/go-spiffe/v2/spiffetls/tlsconfig"
	"github.com/spiffe/go-spiffe/v2/workloadapi"
```

--

While at that, it might be a good idea to move this implementation detail to `spike-sdk-go` too.

Maybe a `ciper` folder here, to follow the same pattern as others (secrets, acl, operator…): https://github.com/spiffe/spike-sdk-go/tree/main/api/internal/impl/api

--

In addition there are certain TLS-level configuration that spike-sdk-go initializes that we skip and assume defaults.

Here's a snippet from the SDK code:

```go
	tlsConfig := tlsconfig.MTLSClientConfig(source, source, authorizer)
	client := &http.Client{
		Transport: &http.Transport{
			TLSClientConfig:     tlsConfig,
			IdleConnTimeout:     30 * time.Second,
			MaxIdleConns:        100,
			MaxConnsPerHost:     10,
			MaxIdleConnsPerHost: 10,
			DialContext: (&net.Dialer{
				Timeout:   30 * time.Second,
				KeepAlive: 30 * time.Second,
			}).DialContext,
			TLSHandshakeTimeout:   10 * time.Second,
			ResponseHeaderTimeout: 10 * time.Second,
			ExpectContinueTimeout: 5 * time.Second,
		},
		Timeout: 60 * time.Second,
	}
```

Better to keep our dirty laundry unified in SDK-landia :) .

app/spike/internal/cmd/cipher/client.go
```diff
@@ -0,0 +1,68 @@
...
+
+	req, err := http.NewRequest("POST", endpoint, in)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/octet-stream")
```
v0lkan: The encrypt/decrypt endpoints also support non-octet-stream payloads.

```
type SecretDecryptRequest struct {
  // Version byte to determine decryption method
  Version byte `json:"version"`
  // Nonce used during encryption
  Nonce []byte `json:"nonce"`
  // Encrypted ciphertext to decrypt
  Ciphertext []byte `json:"ciphertext"`
  // Optional: specify decryption algorithm/version
  Algorithm string `json:"algorithm,omitempty"`
}


// SecretDecryptResponse contains decrypted data
type SecretDecryptResponse struct {
  // Decrypted plaintext data
  Plaintext []byte `json:"plaintext"`
  // Error code if operation failed
  Err data.ErrorCode `json:"err,omitempty"`
}
```

we need to implement an alternative where we pass plaintext etc via command-line flags rather than a file input stream. If the flag exists, we can use it and send the request as JSON; and if not, we can send it as an HTTP stream.

It would be also amazing if the tests cover both use cases.

app/spike/internal/env/net.go
```diff
@@ -0,0 +1,32 @@
...
+
+// NexusBaseURL returns the base URL for SPIKE Nexus.
+// It prefers environment overrides and otherwise falls back to the
+// default in-cluster DNS name used by deployments.
+//
+// Environment variables:
+//   - SPIKE_NEXUS_HOST: scheme and host (e.g., https://spire-spike-nexus.spire-server)
+//   - SPIKE_NEXUS_TLS_PORT: port string (e.g., :8553). If empty, default port is used by server.
+func NexusBaseURL() string {
+	host := os.Getenv("SPIKE_NEXUS_HOST")
```
v0lkan: @parlakisik can you add this to configuration.md and CLAUDE.md?

app/spike/internal/env/net.go
```diff
@@ -0,0 +1,32 @@
...
+	if host == "" {
+		host = "https://spire-spike-nexus.spire-server"
```
v0lkan: for bare-metal deployments without cluster, this will not make sense. 

How will we handle that?

app/spike/internal/env/net.go
```diff
@@ -0,0 +1,32 @@
...
+import "os"
+
+// NexusBaseURL returns the base URL for SPIKE Nexus.
+// It prefers environment overrides and otherwise falls back to the
+// default in-cluster DNS name used by deployments.
+//
+// Environment variables:
+//   - SPIKE_NEXUS_HOST: scheme and host (e.g., https://spire-spike-nexus.spire-server)
+//   - SPIKE_NEXUS_TLS_PORT: port string (e.g., :8553). If empty, default port is used by server.
+func NexusBaseURL() string {
```
v0lkan: @parlakisik where do we use it in the code? --- I couldn't find it.

hack/test/cipher.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
v0lkan, line 1: +1 for adding a test driver.

v0lkan (changes requested):
I have some questions, and a minor documentaiton update request.

Other than that, looks good.

Thanks for your hard work @parlakisik 🙏 .

app/spike/internal/env/net.go
parlakisik: removing this function. it was used in previous implementation without api

app/spike/internal/env/net.go
parlakisik: removed the function

v0lkan:
@parlakisik smoke tests are passing -- we need to test the "actual" functionality; but that will be a task for you :) (i.e. adding it to ./hack/bare-metal/start.sh) -- I'll create an issue, assing it to you, and let you know. -- if that one works too; we can celebrate it with a "beverage" ;)

app/spike/internal/cmd/cipher/decrypt.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/spiffe/spike/app/spike/internal/trust"
+)
+
+func newDecryptCommand(source *workloadapi.X509Source, SPIFFEID string) *cobra.Command {
+	var inFile string
+	var outFile string
+	var versionStr string
+	var nonceB64 string
+	var ciphertextB64 string
+	var algorithm string
```
v0lkan, line 28: is there any reason defining these externally, instead of inside the `RunE` function?

app/spike/internal/cmd/cipher/decrypt.go
```diff
@@ -0,0 +1,122 @@
...
+					if err != nil {
+						return fmt.Errorf("failed to open input file: %w", err)
+					}
+					in = f
+				} else {
+					in = os.Stdin
+				}
+				defer func() {
+					if in != os.Stdin {
+						_ = in.Close()
```
v0lkan, line 51: why are we ignoring the error here?

v0lkan:
Overall, looks good... we can always improve it later.

Crossing fingers and landing this.

PR #214, opened 2025-08-23
remove the caching layer to enable (easier/simpler) HA

v0lkan:
remove the caching layer to enable (easier/simpler) HA


go.mod
```diff
@@ -3,14 +3,15 @@ module github.com/spiffe/spike
 go 1.24.6
 
 require (
+	github.com/DATA-DOG/go-sqlmock v1.5.2
```
v0lkan: instead of SQLmock, we can directly use an SQLite database maybe. -- some of the unit tests already do that. -- need to check why we use a mock in this case.

app/bootstrap/internal/net/net.go
```diff
@@ -77,15 +78,15 @@ func Payload(share secretsharing.Share, keeperID string) []byte {
 		os.Exit(1)
```
v0lkan, line 78: These `os.Exit`s are hard to test -- I'll create an issue to make them testable (by panicking with `log.FatalLn` instead)

app/bootstrap/internal/net/net.go
```diff
@@ -77,15 +78,15 @@ func Payload(share secretsharing.Share, keeperID string) []byte {
...
 	}
 
-	if len(contribution) != 32 {
+	if len(contribution) != crypto.AES256KeySize {
```
v0lkan, line 81: better to use this as a constant instead of hard-coding 32 everywhere.

app/nexus/internal/env/backend.go
```diff
@@ -33,7 +37,7 @@ const (
 // case-insensitive.
 //
 // Valid values are:
-//   - "s3": Uses an AWS S3-compatible medium as a backing store.
```
v0lkan, line 36: No need for an "s3" store type, since we use "lite" in lieu of it.

If we get a request from the community, we can implement one.

app/nexus/internal/initialization/recovery/keeper.go
```diff
@@ -161,22 +117,27 @@ func iterateKeepersAndInitializeState(
...
+			log.FatalLn(fName, "message", "Failed to recover root key")
+		}
 
-		// Both of these methods directly or indirectly make a copy of `binaryRec`
-		// It is okay to zero out `binaryRec` after calling these two functions.
-		state.Initialize(binaryRec)
-		state.SetRootKey(binaryRec)
+		// Both of these methods directly or indirectly make a copy of `rk`
+		// It is okay to zero out `rk` after calling these two functions.
+		state.Initialize(rk)
```
v0lkan, line 129: Moved `SetRootKey` into `Initialize`.

app/nexus/internal/initialization/recovery/recovery.go
```diff
@@ -91,52 +90,13 @@ func InitializeBackingStoreFromKeepers(source *workloadapi.X509Source) {
...
-//	func initializeApp() {
-//		// Other initialization code...
-//		memory.HydrateMemoryFromBackingStore()
-//		// Continue with application startup
-//	}
-func HydrateMemoryFromBackingStore() {
-	const fName = "HydrateMemoryFromBackingStore"
-
-	log.Log().Info(fName, "message", "HydrateMemoryFromBackingStore")
-
```
v0lkan, line 122: we don't need this function anymore since we don't keep state in memoy now.

app/nexus/internal/initialization/recovery/recovery.go
```diff
@@ -260,7 +226,7 @@ func SendShardsPeriodically(source *workloadapi.X509Source) {
 //
 // The function first retrieves the root key from the system state. If no root
 // key exists, it returns an empty slice. Otherwise, it splits the root key into
-// shares using a secret sharing scheme, performs validation checks, and
+// shards using a secret sharing scheme, performs validation checks, and
```
v0lkan, line 229: The docs were using "shares" and "shares" interchangeably; I tried to settle on "shard" as best as I can.

The only place I use "share" is `secretsharing.Share` shares, because the type "is" explicitly named "Share". -- Those shares, are "shards" in our nomenclature, but still, "share" makes it clearer there.

app/nexus/internal/initialization/recovery/recovery.go
```diff
@@ -340,87 +307,3 @@ func NewPilotRecoveryShards() map[int]*[32]byte {
...
-// initialized before keeper distribution to allow immediate operation.
-//
-// Parameters:
-//   - source *workloadapi.X509Source: An X509Source used for authenticating
-//     with keeper nodes
-//
-// The function will crash fatally if:
-//   - Root key creation fails
-//   - The number of keepers doesn't match the configured Shamir shares
-func BootstrapBackingStoreWithNewRootKey(source *workloadapi.X509Source) {
```
v0lkan, line 361: Not using this anymore since we don't bootstrap SPIKE Nexus from within SPIKE Nexus anymore.

app/nexus/internal/initialization/recovery/shamir.go
```diff
@@ -80,6 +81,10 @@ func computeShares() (group.Scalar, []shamir.Share) {
 	defer state.UnlockRootKey()
 	rk := state.RootKeyNoLock()
 
+	if rk == nil || mem.Zeroed32(rk) {
+		log.FatalLn(fName, "message", "root key is nil or zeroed")
+	}
```
v0lkan, line 86: getting a bad root key here is either a configuration, or a security error. It's best to crash loud and clear since this is unexpected.

app/nexus/internal/state/backend/sqlite/ddl/statements.go
```diff
@@ -30,23 +31,27 @@ CREATE TABLE IF NOT EXISTS secrets (
...
 `
 
 // QueryUpdateSecretMetadata is a SQL query for inserting or updating secret
-// metadata. It updates the current version and updated time in conflict with
+// metadata. It updates the current version, oldest version, max versions, and updated time in conflict with
 // the existing path.
 const QueryUpdateSecretMetadata = `
-INSERT INTO secret_metadata (path, current_version, created_time, updated_time)
-VALUES (?, ?, ?, ?)
+INSERT INTO secret_metadata (path, current_version, oldest_version, created_time, updated_time, max_versions)
```
v0lkan, line 48: version information was missing and that was causing test failures.

we didn't realize that before because things were cached in memory; but it was a bug for sure.

app/nexus/internal/state/backend/sqlite/ddl/statements.go
```diff
@@ -93,13 +99,13 @@ WHERE id = ?
 
 // QueryLoadPolicy is a SQL query to select policy details by ID
 const QueryLoadPolicy = `
-SELECT name, spiffe_id_pattern, path_pattern, created_time 
+SELECT name, spiffe_id_pattern, path_pattern, permissions, created_time
```
v0lkan, line 102: We were not persisting permissions either. Major bug.

app/nexus/internal/state/base/path.go
```diff
@@ -22,7 +32,21 @@ package base
...
 func ListKeys() []string {
-	secretStoreMu.Lock()
-	defer secretStoreMu.Unlock()
-	return secretStore.List()
+	ctx := context.Background()
+
+	secrets, err := persist.Backend().LoadAllSecrets(ctx)
+	if err != nil {
+		return []string{}
+	}
```
v0lkan, line 40: the change was involved; but oddly-enough certain methods become much simpler w/o the added caching layer.

app/nexus/internal/state/base/policy.go
```diff
@@ -268,61 +307,22 @@ func ListPoliciesByPath(pathPattern string) []data.Policy {
...
-//   - importedPolicies: A map of policy IDs to policy objects
-//
-// Returns:
-//   - error: An error if any policy fails validation
-func ImportPolicies(importedPolicies map[string]*data.Policy) {
-	for id, policy := range importedPolicies {
-		// Skip nil policies.
-		if policy == nil {
-			continue
-		}
```
v0lkan, line 299: we don't import policies anymore since we fetch them fresh from the db whenever we need it. -- import was to import them into the memory.

docs-src/sass/main.scss
```diff
@@ -51,7 +51,7 @@
 /* System preference based theme */
 @media (prefers-color-scheme: dark) {
   :root:not([data-theme="light"]) {
-    --bg-primary: #1a1a1a;
+    --bg-primary: #101828;
```
v0lkan, line 54: to align with the new landing page theme.

app/nexus/internal/env/backend.go
kfox1111: I think we will eventually want to implement it. but for now, we can remove it.

v0lkan: Yep, makes sense. — I'll create a note for that.

I'll also bump a minor version (instead of a patch version) since this is a relatively bigger change.

Issue #139, opened 2025-05-02
Create a `spike status` command


v0lkan:
create a `spike status` command that shows the current status and
stats of SPIKE Nexus (i.e. whether root key initialized; how many secrets
are in the system; health of SPIKE Nexus and SPIKE Keepers etc.)

It does not have to be too detailed; we can always amend it later.

PR #222, opened 2025-09-02
Implement SPIKE status command for spiffe#139

doguhanniltextra:
This PR implements a new `status` command in the SPIKE CLI.

The command displays basic status and statistics of SPIKE Nexus, including:
- Whether the root key is initialized
- Number of secrets currently stored
- Health status of SPIKE Nexus

This addresses task #139. It provides a simple foundation for checking the system
status. Detailed health checks can be added in future iterations.

### DIFF
--- a/app/nexus/internal/route/health/status.go
+++ b/app/nexus/internal/route/health/status.go
@@ -69,7 +69,7 @@ func RouteGetStatus(
 	status, err := getSystemStatus(r.Context())
 	if err != nil {
 		responseBody := net.MarshalBody(reqres.HealthReadResponse{
-			StatusResponse: data.StatusResponse{},
+			StatusResponse: reqres.StatusResponse{},
 			Err:            data.ErrBadInput,
 		}, w)
 		net.Respond(http.StatusInternalServerError, responseBody, w)
@@ -101,14 +101,14 @@ func RouteGetStatus(
 // goroutines and a context with timeout to prevent blocking forever.
 // Returns a fully aggregated StatusResponse or an error if the operation
 // times out.
-func getSystemStatus(ctx context.Context) (data.StatusResponse, error) {
+func getSystemStatus(ctx context.Context) (reqres.StatusResponse, error) {
 
 	// Set a reasonable timeout for the entire operation
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
 	// Result struct to hold intermediate results
-	res := data.Result{}
+	res := data.HealthResult{}
 
 	var wg sync.WaitGroup
 	var mu sync.Mutex
@@ -153,10 +153,10 @@ func getSystemStatus(ctx context.Context) (data.StatusResponse, error) {
 
 	select {
 	case <-ctx.Done():
-		return data.StatusResponse{}, ctx.Err() // timeout
+		return reqres.StatusResponse{}, ctx.Err() // timeout
 	case <-done:
 		// Return aggregated status
-		return data.StatusResponse{
+		return reqres.StatusResponse{
 			Health:        res.Health,
 			Timestamp:     time.Now(),
 			RootKey:       res.RootKey,
@@ -267,14 +267,14 @@ func rootKeyAvailable() bool {
 }
 
 // getRootKeyStatus checks if the root key is available and returns its status.
-func getRootKeyStatus() data.RootKeyStatus {
+func getRootKeyStatus() data.RootKeyState {
 	if rootKeyAvailable() {
-		return data.RootKeyStatus{
+		return data.RootKeyState{
 			Status: "AVAILABLE",
 		}
 	}
 
-	return data.RootKeyStatus{
+	return data.RootKeyState{
 		Status: "UNAVAILABLE",
 	}
 }
@@ -283,7 +283,7 @@ func getRootKeyStatus() data.RootKeyStatus {
 // It performs a simple write-read-delete operation to verify connectivity
 // and responsiveness. The function returns the backing store status along
 // with its type and response time in milliseconds.
-func getBackingStoreStatus() data.BackingStore {
+func getBackingStoreStatus() data.BackingStoreState {
 	start := time.Now()
 
 	healthy := backingStoreHealthy()
@@ -294,7 +294,7 @@ func getBackingStoreStatus() data.BackingStore {
 		status = "DISCONNECTED"
 	}
 
-	return data.BackingStore{
+	return data.BackingStoreState{
 		Status:         status,
 		Type:           getBackingStoreType(),
 		ResponseTimeMs: &responseTime,
@@ -362,12 +362,12 @@ func getSpireAgentSocketPath() string {
 // getKeeperStatus checks the health of all configured keepers in the cluster.
 // It attempts to connect to each keeper's health endpoint using the provided
 // X.509 source for mTLS authentication. The function counts how many keepers
-func getKeeperStatus(source *workloadapi.X509Source) data.KeeperStatus {
+func getKeeperStatus(source *workloadapi.X509Source) data.KeeperState {
 	keepers := env.Keepers()
 	activeCount := 0
 
 	if len(keepers) == 0 {
-		return data.KeeperStatus{
+		return data.KeeperState{
 			Status:      "NO_KEEPERS_CONFIGURED",
 			ActiveCount: activeCount,
 		}
@@ -405,7 +405,7 @@ func getKeeperStatus(source *workloadapi.X509Source) data.KeeperStatus {
 
 	fmt.Printf("[DEBUG] Keeper status result: %s (active=%d)\n", status, activeCount)
 
-	return data.KeeperStatus{
+	return data.KeeperState{
 		Status:      status,
 		ActiveCount: activeCount,
 	}


app/spike/internal/cmd/status.go
```diff
@@ -0,0 +1,74 @@
...
+			if listSecretKeys != nil {
+				listSecretLength = len(*listSecretKeys)
+			}
+
+			// 5. Determine if the root key is initialized
+			// If there is at least one secret, we assume the root key has been initialized.
+			rootInitialized := false
+			if listSecretLength > 0 {
+				rootInitialized = true
+			}
```
v0lkan: There can be "zero" secrets, but root key may still be initialized.

You cannot deduce existence of the root key merely from secrets length.

app/spike/internal/cmd/status.go
```diff
@@ -0,0 +1,74 @@
...
+			// Retrieve the list of secret keys from the system
+			// ListSecretKeys returns a pointer to a slice of strings. If there is an error,
+			// we cannot determine the system status, so we report an error and exit.
+			listSecretKeys, err := api.ListSecretKeys()
+			if err != nil {
+				fmt.Println("Number of Secrets: ERROR")
+				fmt.Println("SPIKE Nexus Health: ERROR")
+				fmt.Println("SPIKE Keepers Health: ERROR")
+				return err
+			}
```
v0lkan: Instead of making calls here, if we create a `/v1/status` endpoing in **SPIKE Nexus**, we can get a better output as

1. We will be closer to the source of truth.
2. SPIKE Pilot may selectively be deleted/not-deployed, and in that case we won't have anywhere to report the status.
3. Collectors like Prometheus/Grafana cannot use SPIKE Pilot to get status metric.

This CLI interface should call a status API on SPIKE Nexus, and 
1. either display its results
2. or display an error if it cannot talk to the status endpoint.

This way we can tell about
1. the health and status of SPIKE Keepers that talk the SPIKE Nexus
2. the availability of the root key
3. the backing store that SPIKE Nexus uses
4. whether SPIKE nexus is running in FIPS-compliant mode

and many other things, which we cannot reliably tell from an external CLI interface such as SPIKE Pilot.

app/nexus/internal/route/base/impl.go
```diff
@@ -44,6 +44,8 @@ func routeWithBackingStore(a url.APIAction, p url.APIURL) net.Handler {
 		return cipher.RouteEncrypt
 	case a == url.ActionDefault && p == url.NexusCipherDecrypt:
 		return cipher.RouteDecrypt
+	case a == url.ActionGet && p == url.NexusOperatorStatus:
```
v0lkan: "operator" is a special SVID given to an admin who can to doomsday disaster recovery. 

Naming the action as `NexusOperatorStatus` gives the implicaiton that the status can be called only by the operator persona, which is not the case.

The status, for example, may be called from a metrics collector, or a tool like  *SPIKE Bootstrap* before trying to attempt a bootstrap operation. 

the `/status` query is less risky, and it does not make sense to mandate operator privileges to execute it.

I'd use `NexusStatus` or `NexusHealthStatus` and move `status.go` to a `./health` folder from the `./operator` folder too.

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+	"strings"
+
+	"github.com/spiffe/spike-sdk-go/log"
+
+	env "github.com/spiffe/spike/app/nexus/internal/env"
+	state "github.com/spiffe/spike/app/nexus/internal/state/base"
+	"github.com/spiffe/spike/internal/journal"
+
+	"net/http"
+	"time"
```
v0lkan: nit: move these to the top.

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+// BackingStore represents the connection status and performance metrics
+// of the persistent storage backend used for secret storage.
+type BackingStore struct {
+	Status         string `json:"status"`
+	Type           string `json:"type"`
+	ResponseTimeMs *int   `json:"response_time_ms,omitempty"`
+}
+
+// startTime records when the server was started, used for uptime calculation
+var startTime = time.Now()
```
v0lkan: These structs need to go to https://github.com/spiffe/spike-sdk-go

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+
+	if err := json.NewEncoder(w).Encode(status); err != nil {
+		http.Error(w, "Failed to encode status response", http.StatusInternalServerError)
+		return err
+	}
+
+	return nil
+}
```
v0lkan: Please compare and contrast this with `app/nexus/internal/route/secret/get.go` -> `RouteGetSecret`

We use `net.HandleRequest` for an initial sanity check.

Then `guardSecretRequest()` to do things like SPIFFE ID and parameter validation.

We also do a `state.CheckAccess()` in guardsecretrequest to see if there is any policy that explicitly allows this action.

On the topic of CheckAccess; since health status does not have a secret path, you'd want to assign a custom path to that:

```
	allowed := state.CheckAccess(
		sid.String(), "spike/system/status",
		[]data.PolicyPermission{data.PermissionWrite},
	)
```

Then we go out of check access, marshal the request body, and so on.

---

Please make your implementation align with other similar API functions.

You cannot simply check health status without doing ACL verification.

We don't want any random workload in the cluster to hit this endpoint and receive a response.

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+func getSystemStatus() StatusResponse {
+	return StatusResponse{
+		Health:        determineOverallHealth(),
+		Timestamp:     time.Now(),
+		Keepers:       getKeeperStatus(),
+		RootKey:       getRootKeyStatus(),
+		BackingStore:  getBackingStoreStatus(),
+		FIPSMode:      isFIPSMode(),
+		SecretsCount:  getSecretsCount(),
+		UptimeSeconds: int64(time.Since(startTime).Seconds()),
```
v0lkan: Some of these are network-bound and blocking operations.

When these are blocking calls (especially if they involve network requests, database queries, or file I/O), parallelizing them and adding a timeout would be a significant improvement.

Consider creating a wait group, that share a single context, and time out the function if things fail.

```go
func getSystemStatus(ctx context.Context) (StatusResponse, error) {
	// Set a reasonable timeout for the entire operation
	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
	defer cancel()

	type result struct {
		health       HealthStatus
		keepers      KeeperStatus
		rootKey      RootKeyStatus
		backingStore BackingStoreStatus
		fipsMode     bool
		secretsCount int
		err          error
	}

	// Channel to collect results
	ch := make(chan result, 1)

	go func() {
		var res result
		var wg sync.WaitGroup

etc etc
```

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+// Health status levels:
+//   - "OK": All critical components are functioning normally
+//   - "DEGRADED": Backing store issues but system partially functional
+//   - "UNAVAILABLE": Root key unavailable, system cannot perform crypto operations
+//
+// The function prioritizes root key availability over backing store health
+// since cryptographic operations are impossible without the root key.
+func determineOverallHealth() string {
+	// Check all critical components
+	if !isBackingStoreHealthy() {
```
v0lkan: nit. The codebase typically does not use `is` prefix with predicates.

To follow the convention, I'd replace `isBackingStoreHealthy()` with `backingStoreHealthy()`.

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+		return "DEGRADED"
+	}
+
+	if !isRootKeyAvailable() {
+		return "UNAVAILABLE"
```
v0lkan: if the system is "in memory" mode the root key will be unavailable (by design) but the system will be healthy as a horse.

Similarly, in "in memory" mode there won't be any keepers to talk to.

In addition, for the "lite" mode, there won't be any backing store to talk to.

--

I'd also rename "UNAVAILABLE" to "ROOT_KEY_UNAVAIABLE"; it is not clear what is unavailable from the context.

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// determineOverallHealth evaluates all critical system components
+// to determine the overall system health status. The health determination
+// follows a hierarchical approach where certain failures are more critical
+// than others.
+//
+// Health status levels:
+//   - "OK": All critical components are functioning normally
+//   - "DEGRADED": Backing store issues but system partially functional
```
v0lkan: Are you "really" sure when the backing store does not work the system will be functional? — yes, "in theory" things work a little bit, but I wouldn't call it "degraded" as 
- you cannot list secrets
- you cannot create policies
- I don't know if any disaster recovery operations will succeed.

The only viable option would be kill SPIKE Nexus, reboot it, and pray that things go back to normal.

I'd rename the status to "BACKING_STORE_FAILURE"

--

Also note that "in memory" and "lite" modes do not use backing store, so this check does not make sense for either of those.

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+			cleanURLs = append(cleanURLs, trimmed)
+		}
+	}
+
+	activeCount := len(cleanURLs)
+	requiredCountStr := os.Getenv("SPIKE_NEXUS_SHAMIR_THRESHOLD")
+	requiredCount, err := strconv.Atoi(requiredCountStr)
+	if err != nil || requiredCount <= 0 {
+		requiredCount = 2
+	}
```
v0lkan: how do you decide that a Keeper is healthy, without actually making an HTTP mTLS request to it.

If I had `https://example.com:9443` as a keeper url, in the config string, would that mean it is healthy, and well-configured?

This check merely verifies the user's ability to copy and paste.

It's better than nothing; yet we sure can go more.

And the resultant status do not reflect the reality necessarily.

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+	if state.RootKeyZero() {
+		return false
+	}
+
+	keeperStatus := getKeeperStatus()
+	if keeperStatus.ActiveCount < keeperStatus.RequiredCount {
+		return false
+	}
+
+	return true
```
v0lkan: this part does not make sense.

There may me zero keepers, but the root key will still be available in the memory.

Those as disjoint checks.

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+//   - "UNAVAILABLE": Root key missing, zero, or insufficient keepers
+//
+// Source values:
+//   - "keeper": Root key available through keeper reconstruction
+//   - "unknown": Root key source cannot be determined
+func getRootKeyStatus() RootKeyStatus {
+	if isRootKeyAvailable() {
+		return RootKeyStatus{
+			Status: "AVAILABLE",
+			Source: "keeper",
```
v0lkan: From where do you deduce that the source is "keeper"?

Maybe the operator re-seeded the root key via `spike operator recover` and `spike operator restore` without using any keepers. 

How would you pinpoint the source in that case?

More importantly, do you need to provide the source?
Does it add value?

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+func isBackingStoreHealthy() bool {
+	// Defensive health check with panic recovery
+	defer func() {
+		if recover() != nil {
+			// Panic occurred during storage operation
+			// This indicates an unhealthy backing store
+		}
+	}()
+
+	keys := state.ListKeys()
```
v0lkan: ListKey() can be an expensive operation, especially if there is a large amount of secrets.

I'd try setting, getting, and deleting a secret for a more coherent check.

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+// currently configured for the system. This information helps operations
+// teams understand the storage architecture and troubleshoot issues.
+//
+// The function queries the environment configuration to determine the
+// backend type and logs the detected type for debugging purposes.
+//
+// Common backend types include:
+//   - "postgres": PostgreSQL database backend
+//   - "mysql": MySQL database backend
+//   - "file": File system backend
```
v0lkan: Where in the source code did you find the `"file"` store type?

Same goes ith "postgres", and "mysql".

app/nexus/internal/route/operator/status.go
```diff
@@ -0,0 +1,366 @@
...
+//   - *int: Pointer to secret count when backing store is healthy
+//   - nil: When backing store is unhealthy or inaccessible
+func getSecretsCount() *int {
+	secrets := state.ListKeys()
+	if isBackingStoreHealthy() {
+		count := len(secrets)
+		return &count
+	}
+	return nil
+}
```
v0lkan: note that there is no backing store for "in memory" and "lite" modes.

Make sure this function handles that accordingly, or you parse the result properly at the receiving end of this function.

app/keeper/internal/route/store/health.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/spiffe/spike/internal/journal"
+)
+
+type HealthResponse struct {
+	Status        string `json:"status"`
+	MemLocked     bool   `json:"mem_locked"`
+	FIPS          bool   `json:"fips_enabled"`
+	ValidSPIFFEID bool   `json:"valid_spiffe_id"`
+}
```
v0lkan, line 17: Providing FIPS and Memlock status could be a security issue. Nobody needs to know that. If I hit this endpoint and learn that SPIKE Nexus is not FIPS compatible, and it's memory is not locked; as a hacker, it will give me ideas about how to root to the pod and analyze its memory, or whether there are attack vectors that I can leverage on a non-FIPS-compliant Pod — no need to put the image of a watermelon rind to the mind of the donkey, if you know what I mean.

`ValidSPIFFEID` does not make sense. If the SPIFFE ID of the peer is invalid, it should not be able to call the healthcheck endpoint anyway. This is an mTLS connection, both ends validate their peer's SPIFFE IDs, before sending data through.

app/keeper/internal/route/store/health.go
```diff
@@ -0,0 +1,59 @@
...
+	ctx := r.Context()
+	source, selfSPIFFEID, err := spiffe.Source(ctx, spiffe.EndpointSocket())
+	if err != nil {
+		log.Log().Error(fName, "message", "Cannot get SPIFFE source", "error", err)
+		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+		return err
+	}
+	defer spiffe.CloseSource(source)
+
+	validSPIFFE := spiffeid.IsKeeper(env.TrustRoot(), selfSPIFFEID)
```
v0lkan: If SPIFFE ID is not valid, do not return a response saying "hey my SPIFFEID is not valid"

An invalid SPIFFE ID means either there is a configuration problem, or this is a workload trying to impersonate a SPIKE Keeper.

Please check other parts of the codebase about that that control is done.

For example, SPIKE Nexus does not even boot up if its self-spiffe-id is invalid:
https://github.com/spiffe/spike/blob/2244b83a52d8867998b711bac3412c760fb74439/app/nexus/cmd/main.go#L58-L62

app/keeper/internal/route/store/health.go
```diff
@@ -0,0 +1,59 @@
...
+
+	resp := HealthResponse{
+		Status:        "healthy",
+		MemLocked:     mem.Lock(),
+		FIPS:          fips140.Enabled(),
+		ValidSPIFFEID: validSPIFFE,
+	}
+
+	statusCode := http.StatusOK
+	if !resp.MemLocked || !resp.ValidSPIFFEID {
```
v0lkan: memlock is a user option. And there might be reasons FIPS is not enabled/used.
That does not mean SPIKE Keeper is unhealthy.

SPIKE Keeper can be healthy as a mule (albeit less secure) while its memory is not locked, and FIPS disabled (maybe due to a a fips-incompatible system).

---

In fact, SPIKE Keeper is a very basic service that returns a value from its memory. It is very, very, very, VERY hard for it to be in an unhealthy state. And if it is, I'm not sure if it can detect it from within.

Maybe, it's better for SPIKE Nexus to probe its keepers and report their health and reachability.

app/nexus/internal/route/health/status.go
```diff
@@ -0,0 +1,432 @@
...
+)
+
+// startTime records when the server was started, used for uptime calculation
+var startTime = time.Now()
+
+// RouteGetStatus handles GET requests to the /v1/operator/status endpoint.
+// It performs the following steps:
+// Audits the incoming request for monitoring and compliance purposes.
+// Guards the request by validating the SPIFFE ID and checking ACL permissions.
+// Fetches the current system status, including keeper status, root key status, backing store health, FIPS mode, secrets count, and overall health.
```
v0lkan: Nit: Try to keep the documentation and source code within 80-chars length, counting tabs as two characters.

app/nexus/internal/route/health/status.go
```diff
@@ -0,0 +1,432 @@
...
+) error {
+	const fName = "routeGetStatus"
+	// Audit
+	journal.AuditRequest(fName, r, audit, journal.AuditRead)
+
+	requestBody := net.ReadRequestBody(w, r)
+
+	// HandleRequest pattern
+	request := net.HandleRequest[
+		reqres.HealthReadRequest, reqres.HealthReadResponse](
```
v0lkan, line 53: @doguhanniltextra can you add these to SPIKE Go SDK.

It's better to cut a release there, refer them here and then continue review, I think.

Thanks a lot 🙏 .

Please let me know when you create the SDK PR, so I can cut a minor release.

PR #223, opened 2025-09-02
Keep policy information encrypted (spiffe#81)

doguhanniltextra:
Keep policy information encrypted. This PR ensures that all policy data is encrypted at rest and in transit, improving security without affecting existing functionality. Related to spiffe#81.


app/nexus/internal/state/base/policy.go
```diff
@@ -155,6 +156,45 @@ func CreatePolicy(policy data.Policy) (data.Policy, error) {
 		policy.CreatedAt = time.Now()
 	}
 
+	c := persist.Backend().GetCipher()
+	// Get the system cipher from the backend. This cipher holds the key and AEAD logic for encryption.
+
```
v0lkan: Making this change here assumes that we will have a backend that can encrypt at all times—That's not true for "in-memory" backing stores, for example.

A better way to use this functionality is close to where it's used: i.e. in the actual datastore implementation.

Please make your changes at `state/backend/sqlite/persist/policy` level.

the `DataStore` interface already has an `encrypt()` function

```go
		encrypted, nonce, err := s.encrypt(data)
		if err != nil {
			return fmt.Errorf("failed to encrypt secret data: %w", err)
		}
```

See `app/nexus/internal/state/backend/sqlite/persist/secret.go`, for an example.

---

As an aside, The `Backend().GetCipher()` method is meant for the "SPIKE Lite" store to provide encryption as a service; not for direct consumption. So, that's a leaky abstraction and I am taking my notes to rectify it—At least, we should mention where it should (and shouldn't be used) in its documentation, if we cannot fully isolate it.

doguhanniltextra: Thanks for your guidance. I have one question before pushing the new pull request.

 I plan to use the `Datastore` interface for encryption/decryption. However, the `encrypt` method returns `ciphertext`, `nonce`, and `error`:

 ```go
 encryptedSPIFFE, nonceSPIFFE, err := s.encrypt([]byte(policy.SPIFFEIDPattern))
 ```

 Since the nonce is required for decryption, I need to persist both `encryptedSPIFFE` and `nonceSPIFFE`.

 Currently, `QueryUpsertPolicy` looks like this:

 ```sql
 const QueryUpsertPolicy = `
INSERT INTO policies (id, name, spiffe_id_pattern, path_pattern, permissions, created_time)
VALUES (?, ?, ?, ?, ?, ?)
ON CONFLICT(id) DO UPDATE SET
     name = excluded.name,
     spiffe_id_pattern = excluded.spiffe_id_pattern,
     path_pattern = excluded.path_pattern,
     permissions = excluded.permissions
 `
 ```

 To store both values I would need to change the schema and queries (e.g., adding a new column for `nonce`).

My question is: **is modifying the database schema acceptable here, or would you prefer an alternative approach (e.g., encoding `nonce` + `ciphertext` together into the same field, or making encryption optional via a config flag)?**

I want to align with the project’s expectations before making changes that directly affect persistence.

v0lkan: I think a schema change similar to this makes sense:

```go
CREATE TABLE policies (
    id TEXT PRIMARY KEY,
    name TEXT NOT NULL,
    nonce BLOB NOT NULL,  -- Single nonce for all encrypted fields
    encrypted_spiffe_id_pattern BLOB NOT NULL,
    encrypted_path_pattern BLOB NOT NULL,
    encrypted_permissions BLOB NOT NULL,
    created_time INTEGER NOT NULL
);
```

followed by this:

```go
const QueryUpsertPolicy = `
INSERT INTO policies (
    id, 
    name, 
    nonce,
    encrypted_spiffe_id_pattern, 
    encrypted_path_pattern, 
    encrypted_permissions, 
    created_time
)
VALUES (?, ?, ?, ?, ?, ?, ?)
ON CONFLICT(id) DO UPDATE SET
    name = excluded.name,
    nonce = excluded.nonce,
    encrypted_spiffe_id_pattern = excluded.encrypted_spiffe_id_pattern,
    encrypted_path_pattern = excluded.encrypted_path_pattern,
    encrypted_permissions = excluded.encrypted_permissions
`
```

We can modify the database schema. Since the project is at "Developer" maturity stage, we are not concerned about things like backwards compatibility at the moment — But that does not necessarily mean we can go crazy either :D .

app/nexus/internal/state/backend/sqlite/ddl/statement.go
```diff
@@ -12,12 +12,16 @@ const QueryInitialize = `
...
-    spiffe_id_pattern TEXT NOT NULL,
-    path_pattern TEXT NOT NULL,
-    permissions TEXT NOT NULL,
-    created_time DATETIME NOT NULL
+    nonce_spiffe BLOB NOT NULL, 
+    encrypted_spiffe_id_pattern BLOB NOT NULL,
+    nonce_path BLOB NOT NULL,
+    encrypted_path_pattern BLOB NOT NULL,
+    encrypted_permissions TEXT NOT NULL, 
+    created_time INTEGER NOT NULL
```
v0lkan: please use a single nonce; we don't need a separate nonce for each field.

v0lkan (changes requested):

app/nexus/internal/state/backend/sqlite/persist/policy.go
```diff
@@ -103,16 +122,36 @@ func (s *DataStore) StorePolicy(ctx context.Context, policy data.Policy) error {
 		permissionsStr = strings.Join(permissions, ",")
 	}
 
+	// Encryption
+	nonce, err := GenerateCustomNonce(s)
+	if err != nil {
+		return fmt.Errorf("failed to generate nonce: %w", err)
+	}
+	encrypted_spiffe_id, err := EncryptWithCustomNonce(s, nonce, []byte(policy.SPIFFEIDPattern))
```
v0lkan: Can we use `camelCase`; `snake_case` is typically used for languages like Python.

app/nexus/internal/state/backend/sqlite/persist/policy.go
```diff
@@ -156,26 +196,43 @@ func (s *DataStore) LoadPolicy(
...
+	}
+
+	decryptedPermissions, err := s.decrypt(encryptedPermissions, nonce)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decrypt permissions: %w", err)
+	}
+
+	// Set decrypted values
+	policy.SPIFFEIDPattern = string(decryptedSPIFFE)
+	policy.PathPattern = string(decryptedPath)
```
v0lkan: nit: rename as `decryptedSPIFFEIDPattern` `decryptedPathPattern` etc -- pattern "implies" that this is a regex pattern; having clearer variable names will make the intention clear too.

--

For e.g. in context:

path: `tenants/db/credentials`

pathPattern: `^tenants/db/.*$`

app/nexus/internal/state/base/policy.go
```diff
@@ -155,6 +156,45 @@ func CreatePolicy(policy data.Policy) (data.Policy, error) {
...
+	// Encrypt the SPIFFEIDPattern
+	ciphertextID := c.Seal(nil, nonceID, []byte(policy.SPIFFEIDPattern), nil)
+	// Seal encrypts the plaintext with AEAD using the generated nonce.
+
+	policy.SPIFFEIDPattern = string(ciphertextID)
+	// Store the encrypted SPIFFEIDPattern. Nonce must be saved elsewhere if decryption is needed.
+
+	// Generate a nonce for PathPattern
+	noncePath := make([]byte, c.NonceSize())
+	// Create a separate nonce for the path encryption.
```
v0lkan: Looks like we are still using separate nonce's. -- not sure this will work, given we changed the schema to use a single nonce.

app/nexus/internal/state/backend/sqlite/persist/policy.go
```diff
@@ -61,6 +64,22 @@ func (s *DataStore) DeletePolicy(ctx context.Context, id string) error {
...
+}
+
+func EncryptWithCustomNonce(s *DataStore, nonce []byte, data []byte) ([]byte, error) {
+	if len(nonce) != s.Cipher.NonceSize() {
+		return nil, fmt.Errorf("invalid nonce size: got %d, want %d", len(nonce), s.Cipher.NonceSize())
+	}
+	ciphertext := s.Cipher.Seal(nil, nonce, data, nil)
+	return ciphertext, nil
+}
+
```
v0lkan: Why are these methods public? 
we only used them inside the policy package;
suggest making them private.

v0lkan: wrt naming:

* All nonces are "custom" in that they're uniquely generated
* There's no "standard" or "default" nonce that this would be contrasting with
* The implementation is just the standard way to generate and use a nonce

Thus, the prefix "Custom" is redundant here.

Suggest renaming `generateNonce` and `encryptWithNonce`

PR #228, opened 2025-09-05
[DRAFT/NOT-READY] Bootstrap with ConfigMap

v0lkan:
WIP

### DIFF
--- a/app/bootstrap/cmd/main.go
+++ b/app/bootstrap/cmd/main.go
@@ -5,6 +5,7 @@
 package main
 
 import (
+	"context"
 	"crypto/fips140"
 	"flag"
 	"fmt"
@@ -19,10 +20,94 @@ import (
 	"github.com/spiffe/spike/app/bootstrap/internal/net"
 	"github.com/spiffe/spike/app/bootstrap/internal/state"
 	"github.com/spiffe/spike/app/bootstrap/internal/url"
+
+	// Kubernetes client imports
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 )
 
+// shouldSkipBootstrap checks if bootstrap should be skipped based on:
+// 1. Force bootstrap flag
+// 2. Kubernetes ConfigMap state (if running in k8s)
+// 3. Always proceeds in bare-metal environments
+func shouldSkipBootstrap() (bool, string) {
+	const fName = "bootstrap.shouldSkipBootstrap"
+
+	// Check if we're forcing bootstrap
+	if os.Getenv("SPIKE_FORCE_BOOTSTRAP") == "true" {
+		log.Log().Info(fName, "message", "Force bootstrap enabled")
+		return false, ""
+	}
+
+	// Try to detect if we're running in Kubernetes
+	// InClusterConfig looks for:
+	// - KUBERNETES_SERVICE_HOST env var
+	// - /var/run/secrets/kubernetes.io/serviceaccount/token
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		// We're not in Kubernetes (bare-metal scenario)
+		// Bootstrap should proceed in non-k8s environments
+		if err == rest.ErrNotInCluster {
+			log.Log().Info(fName, "message", "Not running in Kubernetes, proceeding with bootstrap")
+			return false, ""
+		}
+		// Some other error - be conservative and proceed
+		log.Log().Warn(fName,
+			"message", "Could not determine cluster config, proceeding with bootstrap",
+			"err", err.Error())
+		return false, ""
+	}
+
+	// We're in Kubernetes - check the ConfigMap
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Log().Error(fName,
+			"message", "Failed to create Kubernetes client, proceeding with bootstrap",
+			"err", err.Error())
+		// Can't check state, proceed with bootstrap to be safe
+		return false, ""
+	}
+
+	// Get the namespace from environment or default to current namespace
+	namespace := os.Getenv("SPIKE_NAMESPACE")
+	if namespace == "" {
+		// Read namespace from service account if not specified
+		if nsBytes, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace"); err == nil {
+			namespace = string(nsBytes)
+		} else {
+			namespace = "spike" // fallback to default
+		}
+	}
+
+	cm, err := clientset.CoreV1().ConfigMaps(namespace).Get(
+		context.Background(),
+		"spike-bootstrap-state",
+		metav1.GetOptions{},
+	)
+	if err != nil {
+		// ConfigMap doesn't exist or can't read it - proceed with bootstrap
+		log.Log().Info(fName,
+			"message", "ConfigMap not found or not readable, proceeding with bootstrap",
+			"err", err.Error())
+		return false, ""
+	}
+
+	skipBootstrap := cm.Data["skip-bootstrap"] == "true"
+	reason := cm.Data["reason"]
+
+	if skipBootstrap {
+		log.Log().Info(fName,
+			"message", "Skipping bootstrap based on ConfigMap state",
+			"reason", reason,
+			"last-check", cm.Data["last-check-time"])
+	}
+
+	return skipBootstrap, reason
+}
+
 func main() {
-	const fName = "boostrap.main"
+	const fName = "bootstrap.main"
 
 	log.Log().Info(fName, "message", "Starting SPIKE bootstrap...", "version", config.BootstrapVersion)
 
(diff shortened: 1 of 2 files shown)


app/bootstrap/cmd/main.go
```diff
@@ -16,28 +16,32 @@ import (
...
+	)
 
 	init := flag.Bool("init", false, "Initialize the bootstrap module")
 	flag.Parse()
 	if !*init {
 		fmt.Println("")
 		fmt.Println("Usage: bootstrap -init")
 		fmt.Println("")
-		os.Exit(1)
+		log.FatalLn(fName, "message", "Invalid command line arguments")
```
v0lkan, line 38: Instead of `os.Exit()` I use FatalLn because FataLn `panic`s so it's easier to write unit tests arond it.

app/bootstrap/cmd/main.go
```diff
@@ -16,28 +16,32 @@ import (
...
 		return
 	}
 
-	// Check if we should skip bootstrap (set by init container)
-	if _, err := os.Stat("/shared/skip-bootstrap"); err == nil {
+	skip := k8s.ShouldSkipBootstrap() // Kubernetes or bare-metal check.
```
v0lkan: maybe use something other than `k8s.` as a namespace since the check applies both for k8s and bare-metal.

app/bootstrap/internal/k8s/k8s.go
```diff
@@ -0,0 +1,193 @@
...
+
+	"github.com/spiffe/spike-sdk-go/config/env"
+	"github.com/spiffe/spike-sdk-go/log"
+	k8s "k8s.io/api/core/v1"
+	k8sMeta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+const k8sTrue = "true"
```
v0lkan: set `"true"` as a constant because we want "exact" case-sensitive match.

That's what I observe in some of the k8s controllers etc: Exact case-sensitive match for booleans.

app/bootstrap/internal/k8s/k8s.go
```diff
@@ -0,0 +1,193 @@
...
+const k8sServiceAccountNamespace = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+
+// ShouldSkipBootstrap determines whether the bootstrap process should be
+// skipped based on the current environment and state. The function follows
+// this decision logic:
+//
+//  1. If SPIKE_BOOTSTRAP_FORCE="true", always proceed (returns false)
```
v0lkan: Need to add this env var to user-facing documentation too.

app/nexus/internal/env/secret.go
```diff
@@ -17,7 +17,7 @@ import (
 // If the environment variable is not set, contains an invalid integer, or
 // specifies a non-positive value, it returns the default of 10 versions.
 func MaxSecretVersions() int {
-	p := os.Getenv(env.NexusMaxSecretVersions)
+	p := os.Getenv(env.NexusMaxEntryVersions)
```
v0lkan, line 20: Linter complained about the name being "insecure" (false positive) -- so I renamed "secret" as "entry".

hack/k8s/Bootstrap.yaml
```diff
@@ -2,6 +2,12 @@
 #  \\\\\ Copyright 2024-present SPIKE contributors.
 # \\\\\\\ SPDX-License-Identifier: Apache-2.0
 
+# TODO: SPIKE_BOOTSTRAP_FORCE should come from SDK constants
+# TODO: update docs that refer SPIKE_BOOTSTRAP_FORCE
+# TODO: update config.md and CLAUDE.md too.
+# TODO: verify that k8s setup work in local k8s.
+# TODO: update the yaml example in the docs too.
```
v0lkan: will address these b4 merge.

app/bootstrap/internal/k8s/k8s.go
```diff
@@ -0,0 +1,193 @@
...
+
+	namespace := "spike"
+	// Read namespace from the service account if not specified
+	if nsBytes, err := os.ReadFile(k8sServiceAccountNamespace); err == nil {
+		namespace = string(nsBytes)
+	}
+
+	cm, err := clientset.CoreV1().ConfigMaps(namespace).Get(
+		context.Background(),
+		"spike-bootstrap-state",
```
kfox1111: Should we make an env var for configmap name?

hack/k8s/Bootstrap.yaml
```diff
@@ -111,9 +96,10 @@ metadata:
   namespace: spike
   name: spiffe-bootstrap-role
 rules:
-- apiGroups: ["batch"]
-  resources: ["jobs"]
-  verbs: ["get", "list"]
+# Allow Bootstrap to read/write ConfigMaps
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get", "create", "update", "patch"]
```
kfox1111: should be able to lock it down slightly more like:
```
- apiGroups: [""]
  resources: ["configmaps"]
  verbs: ["create"]
- apiGroups: [""]
  resources: ["configmaps"]
  verbs: ["get", "update", "patch"]
  resourceNames: ["spike-bootstrap-state"]
```

or, if we change api's slightly, can lock down the create too.

hack/k8s/Bootstrap.yaml
```diff
@@ -62,9 +51,9 @@ spec:
           value: "2"
         - name: SPIKE_NEXUS_KEEPER_PEERS
           value: "https://spiffe-spike-keeper-0.spiffe-spike-keeper-headless:8443,https://spiffe-spike-keeper-1.spiffe-spike-keeper-headless:8443,https://spiffe-spike-keeper-2.spiffe-spike-keeper-headless:8443"
```
parlakisik, line 47: is this static value for peers ?

v0lkan: These will be configurable via Helm charts.

Right now, it's a workaround until we add Bootstrap to https://github.com/spiffe/helm-charts-hardened

hack/k8s/Bootstrap.yaml
```diff
@@ -62,9 +51,9 @@ spec:
           value: "2"
         - name: SPIKE_NEXUS_KEEPER_PEERS
```
v0lkan, line 46: As per @kfox1111 ’s suggestion, I'll rename this to `SPIKE_KEEPER_PEERS`

hack/k8s/Bootstrap.yaml
v0lkan: Yep, makes sense. 

It makes sense not to allow it to R/W "all" configmaps in the NS.

v0lkan: I'll update it accordingly.

docs-src/static/index.html
```diff
@@ -32,7 +32,7 @@
 
       <div class="mx-auto max-w-7xl px-6 pt-10 pb-24 sm:pb-32 lg:flex lg:px-8 lg:py-40">
         <div class="mx-auto max-w-2xl shrink-0 lg:mx-0 lg:pt-8">
-        <img src="/assets/spike-banner.png" alt="SPIKE" class="h-32" />
+        <img src="/assets/spike-banner-dark.png" alt="SPIKE" class="h-32" />
```
v0lkan, line 35: better logo for the darker landing page.

note to self to create a "light" variant of that logo, and swap it on the inner pages based on the theme of the website. — I'll create a "good first issue" for that. — could be fun for those who like to do FrontEnd stuff.

app/bootstrap/internal/k8s/k8s.go
v0lkan: Yep, sound useful. — will do.

app/bootstrap/cmd/main.go
v0lkan: renamed it as "lifecycle": it is a better fit.

hack/k8s/Bootstrap.yaml
v0lkan: Gues I need to rename SPIKE_NEXUS_SHAMIR_SHARES and SPIKE_NEXUS_SHAMIR_THRESHOLD, if I'm following the same logic :) .

v0lkan: Per our follow-up discussion we'll keep the name intact — otherwise we'll be introducing a bunch of changes to the code (and helm charts) to keep things backwards-compatible.

app/bootstrap/internal/lifecycle/lifecycle.go
```diff
@@ -0,0 +1,193 @@
...
+//
+//  1. If SPIKE_BOOTSTRAP_FORCE="true", always proceed (returns false)
+//  2. In bare-metal environments (non-Kubernetes), always proceed
+//  3. In Kubernetes environments, check the "spike-bootstrap-state" ConfigMap:
+//     - If ConfigMap exists and bootstrap-completed="true", skip bootstrap
+//     - Otherwise, proceed with bootstrap
+//
+// The function returns true if bootstrap should be skipped, false if it
+// should proceed.
+func ShouldSkipBootstrap() bool {
```
v0lkan: I think it's more intuitive to invert the logic and call this `ShouldBootstrap()` :) — I end up reading things twice every time I hit a `return false` :D .

docs-src/content/getting-started/quickstart.md
```diff
@@ -8,6 +8,19 @@ weight = 2
...
+> **⚠️ This Page Needs Work**
+> 
+> Since **SPIFFE Helm Charts** do not have **SPIKE Bootstrap** yet, the 
+> instructions on this page are incomplete and will likely fail.
+> 
+> If you want to test **SPIKE** on a Kubernetes cluster, we recommend
+> using the [**SPIKE Quickstart Guide**][quickstart] instead.
+> 
+> We will update this page once **SPIKE Bootstrap** is available in the 
+> upstream **SPIFFE Helm Charts**.
```
v0lkan, line 20: @kfox1111 FYI -- I'll let you know when we cut a release so that we can reference Bootstrap in charts.

PR #239, opened 2025-10-09
Bootstrap PoP validation

v0lkan:
And various other changes... I will annotate the code.

### DIFF
--- a/app/bootstrap/cmd/main.go
+++ b/app/bootstrap/cmd/main.go
@@ -72,11 +72,6 @@ func main() {
 	// Retries verification until successful.
 	net.VerifyInitialization(ctx, api)
 
-	// TODO: both verification and retry is done forever; we need an upper limit
-	// for these.
-	// we can have a goroutine/channel that panic if this entire main thread
-	// does not complete within a given timeframe.
-
 	// Bootstrap verification is complete. Mark bootstrap as "done".
 
 	// Mark completion in Kubernetes
--- a/jira.xml
+++ b/jira.xml
@@ -41,6 +41,12 @@
       guide still works.
     </issue>
     <issue>
+      // TO: both verification and retry is done forever; we need an upper limit
+      // for these.
+      // we can have a goroutine/channel that panic if this entire main thread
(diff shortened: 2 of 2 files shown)


app/bootstrap/cmd/main.go
```diff
@@ -80,15 +87,15 @@ func main() {
 
 	ctx := context.Background()
 
-	for keeperID, keeperAPIRoot := range env.Keepers() {
+	for keeperID, keeperAPIRoot := range env.KeepersVal() {
```
v0lkan: Getting the environment values from the SDK, since most of them are being commonly shared across various SPIKE components.

In addition, having them all together in the SDK doubles as documentation too.

app/bootstrap/cmd/main.go
```diff
@@ -105,20 +112,107 @@ func main() {
...
 		if err != nil {
 			log.FatalLn(fName, "message", "Initialization failed", "err", err)
 		}
-
 	}
 
 	log.Log().Info(fName, "message", "Sent shards to SPIKE Keeper instances.")
 
+	// Verify that SPIKE Nexus has been properly initialized by sending an
+	// encrypted payload and verifying the hash of the decrypted plaintext.
```
v0lkan, line 71: I will double check that this works before merging this PR.

app/bootstrap/cmd/main.go
```diff
@@ -51,73 +45,39 @@ func main() {
...
-			return true, nil
-		},
-			retry.WithBackOffOptions(
-				retry.WithMaxInterval(60*time.Second), // TODO: to env vars.
-				retry.WithMaxElapsedTime(0),           // Retry forever.
-			),
-		)
+	// Broadcast shards to the SPIKE keepers until all shards are
+	// dispatched successfully.
+	net.BroadcastKeepers(ctx, api)
```
v0lkan, line 66: This one, and the verification that follows it run forever...
We will need to have a "bail out" timeout as an env var; but that can be done as a follow-up PR.

PR #246, opened 2025-11-24
Refactor data dir config and improve test isolation

ewriq:
- The old withEnvironment is gone; We use t.Setenv.
- sqlite/lite tests are now in their own directory (SPIKE_NEXUS_DATA_DIR) with t.TempDir.
- init_sqlite.go now calls SpikeNexusDataFolder() (PathProvider will be installed in the future).
- Organized a few tests, shared db race is decreasing.


ewriq:
Thanks for your comments 😊🙏

PR #248, opened 2025-12-08
Documentation Update

v0lkan:
Updating documentation to reflect the new architecture, replacing older images, etc.


v0lkan:
The linter error is a false positive; I'll just ignore it. — There is no patch in the referenced code that can call an x.509 certificate read operation.

PR #260, opened 2025-12-12
(feat) add retry and fail logic with configurable timeout and maximum…

ozkanturidi:
(feat) add retry and fail logic for bootstrap keeper with configurable timeout and maximum number of retries variables


app/bootstrap/internal/net/broadcast.go
```diff
@@ -28,10 +29,16 @@ import (
 // BroadcastKeepers distributes root key shares to all configured SPIKE Keeper
 // instances. It iterates through each keeper ID from the environment
 // configuration and sends the corresponding keeper share using the provided
-// API. The function retries indefinitely until each share is successfully
-// delivered. If a keeper fails to receive its share, the function logs a
-// warning and retries. The function terminates the application if the retry
-// mechanism fails unexpectedly.
+// API. The function retries retries with exponential backoff using a configurable timeout and a maximum
+// number of retry attempts per keeper. If a keeper cannot be reached within the timeout, the
```
v0lkan: nit: can we limit the line lenght to 80 chars pls.

app/bootstrap/internal/net/broadcast.go
```diff
@@ -40,34 +47,44 @@ func BroadcastKeepers(ctx context.Context, api *spike.API) {
...
+
+		_, err := retry.WithMaxAttempts(keeperCtx, maxRetries,
+			func() (bool, *sdkErrors.SDKError) {
+				log.Info(fName,
+					"message", "sending shard to keeper",
+					"keeper_id", keeperID,
+					"keeper_url", keeperURL,
+				)
+
+				err := api.Contribute(keeperShare, keeperID)
```
v0lkan: If this request hangs:

1. The keeperCtx deadline expires after timeout
2. The context's Done() channel gets closed
3. Nothing happens — because nothing is listening to that channel
4. api.Contribute continues blocking forever

Suggestion: wrap this in a goroutine:

```
func contributeWithContext(ctx context.Context, api *spike.API, share []byte, keeperID string) error {
    done := make(chan error, 1)
    
    go func() {
        err := api.Contribute(share, keeperID)
        done <- err
    }()
    
    select {
    case err := <-done:
        return err
    case <-ctx.Done():
        return ctx.Err()
    }
}
```

and

```
err := contributeWithContext(keeperCtx, api, keeperShare, keeperID)
```

app/bootstrap/internal/net/broadcast.go
```diff
@@ -182,3 +203,28 @@ func AcquireSource() *workloadapi.X509Source {
...
+		done <- err
+	}()
+
+	select {
+	case err := <-done:
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
```
v0lkan, line 230: This is obviously hacky; but a "proper" solution will require the refactoring beaver to add context to all API methods :D -- so for now, this is the best compromise :)

Issue #249, opened 2025-12-09
`spike policy list` should have concise output


v0lkan:
`spike secret list` and `spike policy list` have inconsistent output behavior. The secret list command shows only paths (concise), while the policy list command dumps entire policy objects (verbose). 

For consistency and usability, spike policy list should show only policy names and IDs.

Current Behavior:

`spike secret list` - Concise output:

```
secrets/db/creds
secrets/app/config
secrets/cache/redis
```

`spike policy list` - Verbose output (dumps entire policy objects):

``` 
- name: "database-policy"
  spiffeidPattern: "^spiffe://example\\.org/database$"
  pathPattern: "^secrets/database/production$"
  permissions: [read]
- name: "web-service-policy"
  spiffeidPattern: "^spiffe://example\\.org/web-service$"
  pathPattern: "^secrets/web-service/config$"
  permissions: [read, write]
...
```

Expected Behavior

spike policy list should show only names and IDs:

```
ID                                    NAME
550e8400-e29b-41d4-a716-446655440000  database-policy
6ba7b810-9dad-11d1-80b4-00c04fd430c8  web-service-policy
```

Users can then use spike policy get <name-or-id> to retrieve full policy details when needed.

Rationale

1. Consistency: Both list commands should follow the same pattern (concise summary)
2. Usability: Policy lists can grow large; dumping full objects makes output hard to scan
3. Separation of concerns: List operations should enumerate resources; get operations should show details
4. CLI conventions: Standard practice is for list to show identifiers and get/describe to show details

PR #263, opened 2025-12-16
feat(policy): render spike policy list output as ID/NAME table

yagmurcicekdagi:
Fixes: #249

- Return `PolicyListItem` from Nexus policy list route
- Format CLI list output with tabwriter header + rows
- Update tests to use a normalized-output helper function
- Adjust `start.sh `validation to match the new table format


v0lkan (changes requested):
@yagmurcicekdagi can you please resolve the conflicts?

It might be possible that `main` and your changes might be trying to fix the same thing.

Feel free to carry your changes over, or pick from `main` depending on whichever makes sense.

Make sure you run `make build`; `make test`; `make audit`; and `make start` locally to ensure things are in good shape.

Thanks a lot 🙏 .

Issue #251, opened 2025-12-09
Encrypt secret metadata at rest (currently only secret values are encrypted)


v0lkan:
SPIKE currently encrypts secret values but stores secret metadata in plaintext. 

In contrast HashiCorp Vault encrypts all secret metadata, leaving only the path for indexing. 

SPIKE should follow the same approach for defense in depth, and also to provide similar UX for people who are accustomed to Vault.

Current Behavior

In the SQLite backing store:
- Encrypted: Secret values
- Plaintext: Secret metadata (created time, updated time, version info, etc.)

Proposed Behavior

Encrypt all secret metadata at rest, similar to how policies are handled:
- Encrypted: Secret values AND metadata
- Plaintext: Only the path (required for indexing/lookup)

Rationale

1. Defense in depth: If an attacker gains access to the database file, metadata exposure reveals:
  - When secrets were created/modified (operational patterns)
  - How frequently secrets rotate (security posture hints)
  - Number of versions retained (retention policy details)
2. Vault parity: Vault encrypts metadata; users migrating from Vault expect equivalent protection
3. Consistency: SPIKE already encrypts policy metadata (spiffeid pattern, path pattern, permissions). Secret metadata should receive the same treatment
4. Compliance: Some security frameworks require encryption of all sensitive operational data, not just the secrets themselves

Implementation Notes

- The path must remain in plaintext for indexing and lookup operations
- Follow the same encryption pattern used for policies in app/nexus/internal/state/backend/sqlite/persist/policy.go
- Schema change required in app/nexus/internal/state/backend/sqlite/ddl/statement.go

PR #277, opened 2026-01-01
feat(secret): encrypt secret metadata at rest

yagmurcicekdagi:
Fixes: #251

- Encrypt secret metadata at rest by storing encrypted fields and a metadata nonce in secret_metadata.
- Update SQL schema and queries to handle encrypted metadata columns and nonce, with correct insert/update semantics.
- Decrypt metadata on load and decode fields back to ints/timestamps.
- Adjust tests/helpers to insert encrypted metadata

### DIFF
--- a/app/nexus/internal/state/backend/sqlite/persist/field_nonce.go
+++ b/app/nexus/internal/state/backend/sqlite/persist/field_nonce.go
@@ -1,0 +1,118 @@
+//    \\ SPIKE: Secure your secrets with SPIFFE. — https://spike.ist/
+//  \\\\\ Copyright 2024-present SPIKE contributors.
+// \\\\\\\ SPDX-License-Identifier: Apache-2.0
+
+package persist
+
+import (
+	"fmt"
+
+	sdkErrors "github.com/spiffe/spike-sdk-go/errors"
+)
+
+const (
+	nonceFieldSecretMetadataCurrentVersion string = "secret_metadata.current_version"
+	nonceFieldSecretMetadataOldestVersion  string = "secret_metadata.oldest_version"
+	nonceFieldSecretMetadataCreatedTime    string = "secret_metadata.created_time"
+	nonceFieldSecretMetadataUpdatedTime    string = "secret_metadata.updated_time"
+	nonceFieldSecretMetadataMaxVersions    string = "secret_metadata.max_versions"
+)
+
+// fieldNonceSalts are fixed per-field salts (must match AES-GCM nonce size).
+// A per-field nonce is derived as baseNonce XOR salt.
+var fieldNonceSalts = map[string][]byte{
+	nonceFieldSecretMetadataCurrentVersion: []byte("current_ver_"), // 12 bytes
+	nonceFieldSecretMetadataOldestVersion:  []byte("oldest_ver__"), // 12 bytes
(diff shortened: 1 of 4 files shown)


v0lkan:
Thanks for the contribution 🙏 -- there is some upstream refactoring in progress; I'll check this immediately after that.

app/nexus/internal/state/backend/sqlite/ddl/statement.go
```diff
@@ -76,15 +79,15 @@ ON CONFLICT(path, version) DO UPDATE SET
 
 // QuerySecretMetadata is a SQL query to fetch metadata of a secret by its path.
 const QuerySecretMetadata = `
-SELECT current_version, oldest_version, created_time, updated_time, max_versions
-FROM secret_metadata 
+SELECT nonce, encrypted_current_version, encrypted_oldest_version, encrypted_created_time, encrypted_updated_time, encrypted_max_versions
```
v0lkan, line 82: Using a single nonce per field would be cryptographically weak and it can cause warnings in security audits.

While I do realize that the data here itself is not super-sensitive, auditors typically want to "check boxes", rather than having a pragmatic perspective on security.

My suggestion is as follows:

Derive per-field nonces from the stored base nonce without changing the schema:

```
  var fieldNonceSalts = map[string][]byte{
      "current_version": []byte("current_ver_"),  // 12 bytes each
      "oldest_version":  []byte("oldest_ver__"),
      "created_time":    []byte("created_tim_"),
      // ...
  }

  func deriveFieldNonce(baseNonce []byte, field string) []byte {
      derived := make([]byte, len(baseNonce))
      salt := fieldNonceSalts[field]
      for i := range baseNonce {
          derived[i] = baseNonce[i] ^ salt[i]
      }
      return derived
  }
```
                                                                                                                                                                                                              This gives each field a unique nonce while keeping a single nonce column in the DB. Decryption remains deterministic since the derivation is reversible.

---

One related question would be "what if an attacker knows the per-field key" (since they have access to the source code, as the code is public).
                                                                                                                                                                                                              Kerckhoffs's principle states that "A cryptosystem should be secure even if everything except the key is public knowledge".

In AES-GCM:
- Key → must be secret (this is your root key)
- Nonce → must be unique per encryption, but can be public

The nonce is typically stored in plaintext right next to the ciphertext (which we are already doing). An attacker seeing the nonce doesn't help them—they still can't decrypt without the key.

The per-field salts being in the source code is fine because:
1. The base nonce is random (generated via crypto/rand)
2. XORing with known constants just ensures each field's nonce differs
3. The actual security comes entirely from the encryption key

As an aside, nonces don't need to be secret—they only need to be unique.

yagmurcicekdagi: Thanks for the review! 

My initial thought was to generate unique nonces per field, but as you mentioned, this would be an overkill, so I opted for the single nonce approach—referring to how policies are encrypted in app/nexus/internal/state/backend/sqlite/persist/policy.go.

I'm also just curious: Is there a specific reason why single nonce per field approach is used in policies? Wouldn't that be cryptographically weak as well?

v0lkan: > I'm also just curious: Is there a specific reason why single nonce per field approach is used in policies?

It's a pragmatic decision. A policy is relatively less sensitive than a secret.

PR #284, opened 2026-02-09
feat: simplify policy identification by using name as primary identif…

Aliexe-code:
Summary
This PR removes the auto-generated UUID ID field from SPIKE policies and uses the user-provided name as the sole unique identifier, aligning with HashiCorp Vault's approach for better UX and simplicity.

Problem
SPIKE policies had two identifiers:
- ID: Auto-generated UUID (e.g., 550e8400-e29b-41d4-a716-446655440000)
- name: User-provided name (e.g., database-policy)

Testing
- All existing tests pass
- Build succeeds
- No breaking changes to existing behavior (IDs are internal only) 

 - This change is backward compatible for end users since they always used name in CLI commands. The ID field was only used internally and never exposed to users in the API responses of these endpoints.
Follow-up Required ⚠️

SDK Dependency: The spike-sdk-go package contains request/response structs that still use ID fields. This PR includes a temporary workaround using the SDK's ID field but treating it as the policy name. A follow-up PR to the SDK is needed to:
- Rename PolicyReadRequest.ID to PolicyReadRequest.Name
- Rename PolicyDeleteRequest.ID to PolicyDeleteRequest.Name
- Update PolicyListItem.ID to be consistent

- issue #250

Issue #250, opened 2025-12-09
Simplify policy identification: use name as primary identifier; remove ID from policies


v0lkan:
SPIKE policies currently have both an ID (UUID) and a name field, which adds unnecessary complexity. 

In comparison, HashiCorp Vault policies use only a name as the unique identifier. Following the same approach will provide simplicity and better UX.

Current Behavior

SPIKE policies have two identifiers:
- ID: Auto-generated UUID (e.g., 550e8400-e29b-41d4-a716-446655440000)
- name: User-provided name (e.g., database-policy)

This creates ambiguity:
- Which identifier should users use for spike policy get?
- Which identifier appears in logs and audit trails?
- Users must track both values or look up one from the other

Expectation:

Use name as the sole unique identifier for policies.
Remove the ID field from db schema.

```
# Create policy with unique name
spike policy create database-policy --spiffeid-pattern "..." --path-pattern "..."

# Reference by name only
spike policy get database-policy
spike policy delete database-policy
```

Rationale

1. Simplicity: One identifier is easier to remember and use than two
2. Vault compatibility: Users familiar with Vault will expect this pattern
3. Human-readable: Names are meaningful; UUIDs are not
4. Fewer moving parts: No need to generate, store, or index UUIDs
5. Better UX: spike policy get database-policy is clearer than spike policy get 550e8400-e29b-41d4-a716-446655440000

PR #292, opened 2026-07-14
feat(policy): use name as the sole policy identifier

v0lkan:
Folds in the substantive work from PR #284 (issue #250) onto current main, resolving the drift from the spike-sdk-go v0.19.9 migration.

SPIKE policies are uniquely identified by their user-provided name, so the auto-generated UUID id added drift and a practically unused column. This removes it end to end:

- Drop the id column from the policies table; name is now the PRIMARY KEY. Upsert, load, and delete queries all key on name.
- Persist, state, memory backend, and CLI operate on name; delete and load-by-name replace the id-based paths.
- google/uuid is no longer imported directly (demoted to indirect).

Conflict resolutions preserve main's post-migration idioms:
- persist StorePolicy/DeletePolicy keep the withSerializableTx helper while switching to name-based args
- CLI keeps the context.Context threading; findPolicyByName uses the ctx-aware ListAllPolicies

The SDK still exposes the identifier as PolicyReadRequest.ID and PolicyDeleteRequest.ID, so request.ID carries the policy name at the route boundary. This bridge is documented inline; issue #250 tracks the pending SDK field rename. (TODO markers from the original PR were converted to plain comments per the no-TODO-in-main rule.)

make test and make audit both pass; govulncheck reports 0 affecting vulnerabilities.

cc: @Aliexe-code

Spec: TBD


v0lkan:
Supersedes and closes: https://github.com/spiffe/spike/pull/284

v0lkan:
reminder. Issue #250 is still open and it is related to this and possibly the next thing to be done.

